PR #18, opened 2015-10-28
vasp.jobs: catch already encoded default_vis

tschaume:
see subject line. There might be a better (non-temporary) fix for this.

### DIFF
--- a/custodian/vasp/jobs.py
+++ b/custodian/vasp/jobs.py
@@ -290,10 +290,14 @@ class VaspJob(Job):
                           settings_override=settings)
 
     def as_dict(self):
+	try:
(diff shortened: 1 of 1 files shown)


shyuep:
This should not be the job of VaspJob. The fundamental problem is with fireworks changes introduced by @computron to automatically decode. That should be fixed on the workflow end of things.

In general, try to decode an already decoded object should always result in errors.

computron:
Patrick, I did not review your commit carefully in our other (offline)
message (i.e., I did not correctly read which lines were added and
subtracted in that email, which did not contain formatting/colors). Seeing
it on git, I agree with Shyue that the proposed changes should not be added
and the problem should be fixed upstream.

PR #21, opened 2016-02-01
alternative strategy to avoid BRMIX error

tschaume:
Our current VASP installation (v5.2) used for MP production was compiled using the Intel compiler. Apparently, this causes some of our static GGA calculations to throw a serious BRMIX error, which results in custodian to reach its maximum allowed errors. After quite some trial and error, the following strategy using a combination of generation style, kpoints and ISYM settings seems to allow many of our failed static GGA runs to complete successfully.

For instance, starting with the default settings of

```
Monkhorst   —   [12, 12, 12]   —   ISYM default
```

the static GGA run throws a BRMIX error and custodian aborts it after a certain step - as expected. Normally, custodian would now react by setting ISYM to 0, changing to the Gamma generation style, and removing the CHGCAR/WAVECAR files. This will cause the job to restart another 4 times until it hits the maximum number of errors allowed and then fizzle. However, I can get the run to complete successfully on the second try, by instead keeping the CHGCAR/WAVECAR files and the ISYM default, switching to Gamma generation style, and increasing the kpoints by one to make it odd:

```
keep CHGCAR and WAVECAR
Gamma       —   [13, 13, 13]   —   ISYM default
```

The only caveat is the warning

```
random wavefunctions but no delay for mixing, default for NELMDL
```

where `initial charge from wave function` should show up in the VASP output.

This pull request is a preliminary implementation of this strategy. It’s obviously conflicting with what custodian would normally do in this situation. @shyuep, what would be a good way to implement multiple different recipes to be tried for the same error - in this case, the above strategy before switching to ISYM = 0 (which apparently runs significantly slower)?

Note that this implementation is currently being used in MP's production but does not necessarily need to be part of the official custodian. I'm fine with simply keeping this pull request and the respective branch for record and discussion purposes only.

### DIFF
--- a/custodian/vasp/handlers.py
+++ b/custodian/vasp/handlers.py
@@ -22,6 +22,7 @@ import datetime
 import operator
 import shutil
 from functools import reduce
+from collections import Counter
 import re
 
 from six.moves import map
@@ -96,6 +97,7 @@ class VaspErrorHandler(ErrorHandler):
         """
         self.output_filename = output_filename
         self.errors = set()
+        self.error_count = Counter()
 
     def check(self):
         incar = Incar.from_file("INCAR")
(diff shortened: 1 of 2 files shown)


shyuep:
I would say that the strategy should just account for the number of errors that have occurred. If the BRMIX has occurred more than once, then the new strategy would be used. This can be done by introducing a state variable, e.g., self.error_count["brmix"] = 1, 2, ....

tschaume:
Ok, thanks. I added a commit in which custodian employs different strategies based on a counter for BRMIX errors. It's been successfully tested with a few static GGA runs in MP production. If the current solution is OK, I can also fix the unit test if you plan to merge this pull request.

shyuep:
Looks good. Go ahead.

tschaume:
The unit test for the BRMIX error handler is fixed. I included a more meaningful test to assure that the number of kpoints is odd after calling `h.correct()`.

shyuep:
Thanks for the quick response!

PR #22, opened 2016-02-08
Added **kwargs to all Vasp Error Handler constructors to prevent crashes

dcossey014:
Dr. Shyue Ping Ong and Dr. Anubhav Jain,

This change will prevent handlers from crashing when altering parameters from multiple handlers using the following:

def load_class(mod, name):
    mod = __import__(mod, globals(), locals(), [name], 0)
    return getattr(mod, name)
handlers = [load_class("custodian.vasp.handlers", n)(**self.get('handler_params', {})) for n in hnames]

If altering parameters for more than one handler, handler constructors will crash when seeing parameters not in their __init__ without **kwargs.  

Regards,
David Cossey

### DIFF
--- a/custodian/vasp/handlers.py
+++ b/custodian/vasp/handlers.py
@@ -85,7 +85,7 @@ class VaspErrorHandler(ErrorHandler):
         "edddav": ["Error EDDDAV: Call to ZHEGV failed"]
     }
 
-    def __init__(self, output_filename="vasp.out"):
+    def __init__(self, output_filename="vasp.out", **kwargs):
         """
         Initializes the handler with the output file to check.
 
@@ -299,7 +299,7 @@ class AliasingErrorHandler(ErrorHandler):
                            "for an accurate"]
     }
 
-    def __init__(self, output_filename="vasp.out"):
+    def __init__(self, output_filename="vasp.out", **kwargs):
         """
         Initializes the handler with the output file to check.
 
@@ -379,7 +379,7 @@ class MeshSymmetryErrorHandler(ErrorHandler):
     is_monitor = False
 
     def __init__(self, output_filename="vasp.out",
-                 output_vasprun="vasprun.xml"):
+                 output_vasprun="vasprun.xml", **kwargs):
         """
         Initializes the handler with the output files to check.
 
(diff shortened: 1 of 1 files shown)


shyuep:
I am not quite sure what's the use case here. We do want a bad deserialization to crash. Adding **kwargs allows a constructor to take any kind of argument, which is very very bad unless it is a generic use class like dict.

dcossey014:
I have been reading up on the use/abuse of **kwargs and I understand your misgivings about implementing it.  Using the pymatgen.io.vasp.interfaces.py classes, we have a vasp_interface_defaults.yaml file similar to the InputSets YAML files.  This file allows us to easily customize various parameters like KPOINTS settings and Custodian Parameters without scripting in each change to every python script (see below for example).  We were using this to modify a couple different handler parameters (FrozenJobErrorHandler timeout and MaxForceErrorHandler threshold) for HSE and optical properties.  When running Optical calculations, there were long periods (especially with high KPOINT grids) where vasp.out would not be written to and the simulation would be terminated.  This was one of the biggest desires for implementing it into the Defaults File.  We wanted to keep using the FrozenJob handler to prevent wasting CPU hours, but needed to give VASP enough time to run the Optical Calculations.

But our function for loading handlers for Custodian was copied from fireworks_vasp.scripts and it was limited in how it fed handler parameters down to the handlers.  I'll see if I can work more logic into that function to eliminate the need for **kwargs in the handlers.

Defaults File:
VASP_INPUT_SET:  MinimalVaspInputSet
INPUT_SET_PARAMS:
  user_incar_settings:
    ALGO: Normal
    LOPTICS: .TRUE.
    CSHIFT: 0.1 
    NEDOS: 4096
    SIGMA: 0.01
CUSTOM_PARAMS:
  user_kpts_settings:
    kpts_style:  Gamma
    kpts: 
    -   16
    -   16
    -   16
    kpts_shift:  
    -   0
    -   0
    -   0
CUSTODIAN_PARAMS:
  handlers:
    -   MaxForceErrorHandler
    -   FrozenJobErrorHandler
  handler_params:
    timeout: 10800
    max_force_threshold: 0.20

PR #28, opened 2016-06-10
minor fix to vis for deserialization

montoyjh:
Minor fix for jobs.py to ensure deserialization.  Note that it causes a warning to be issued from using deprecated MPVaspInputSet.


tschaume:
please see #18 for the same (but rejected) pull request and related discussion.

PR #38, opened 2016-12-15
Error on runcode

shyamd:
Makes Custodian throw an error if the Popen subprocess terminates with a nonzero returncode. Should ensure custodain fails properly if the error handlers don't already catch this.


shyuep:
Pls implement test. Given that this is such a critical issue.

PR #39, opened 2016-12-15
Added Test for nonzero return code

shyamd:
Added a test to ensure custodian fails on nonzero return code from subprocesses

### DIFF
--- a/custodian/tests/test_custodian.py
+++ b/custodian/tests/test_custodian.py
@@ -20,7 +20,21 @@ from custodian.custodian import Job, ErrorHandler, Custodian, Validator
 import os
 import glob
 import shutil
+import subprocess
 
+class ExitCodeJob(Job):
+
+    def __init__(self, exitcode = 0):
+        self.exitcode = exitcode
+
+    def setup(self):
+        pass
+
+    def run(self):
+        return subprocess.Popen('exit {}'.format(self.exitcode),shell=True)
+
+    def postprocess(self):
+        pass
 
 class ExampleJob(Job):
 
@@ -196,6 +210,13 @@ custodian_params:
         self.assertEqual(len(c.jobs), 2)
         self.assertEqual(len(c.handlers), 3)
         self.assertEqual(len(c.validators), 1)
+
+    def test_exitcode_error(self):
+        c = Custodian([],[ExitCodeJob(0)])
+        output = c.run()
+
+        c = Custodian([],[ExitCodeJob(1)])
+        self.assertRaises(RuntimeError, c.run)
 
     def tearDown(self):
         for f in glob.glob("custodian.*.tar.gz"):


shyuep:
Thanks.

xhqu1981:
@shyuep @shyamd What are the return codes of a recoverable error? If I understand correctly, CustodianError is for unrecoverable errors. That means, if any one of the recoverable errors also has a nonzero return code, custodian will just bypass the handler and hence loss the opportunity to fix it.

computron:
@shyamd @xhqu1981 

I am not sure about VASP,  but certainly in the most general case I can see someone wanting to use a handler to correct a non-zero error code. In that sense, immediately throwing an exception might not be best way forward, but at least custodian should signal that the run has an error. Another option is to implement a Validator that you can easily set when you want to make sure the final thing (after all jobs and handlers are done) doesn't have non-zero returncode rather than try to catch this after the job step itself.

xhqu1981:
agree

shyuep:
Just to be clear - Custodian's job is not to catch error codes. IN the first place, these are not well-defined. 

If a job is bad (whether it is an error code, or bad outputs), that should be dealt with by either a Handler or a Validator. I have no idea what kind of crazy VASP NERSC uses that causes the non-zero error codes, but if you need to check the vasp.out, go ahead and implement a Handler to check those.

I will now revert this pull req.

shyamd:
Wait, why are we reverting this? It's good default behavior if your job returns a Popen object and if you want to do something with the return code then your job doesn't return a Popen object: problem solved.

shyuep:
I am not saying a non-zero return code should be ignored. If you need to handle it, it should be done via an implemented Handler or Validator, not by exiting Custodian with an Exception. In most instances with a non-zero code, the output will be gibberish and the validator should already catch it. I do not know what happens on NERSC, but if there are problems taht do not result in the validators catching them, by all means implement a new validator to catch those cases.

I have added a warning with a non-zero return code.

shyamd:
Given that there is no way to pass live variables to either error handlers or validators, how do you think this should be done?

computron:
I think it would be nice if a Validator could access the returncode rather than just the directory output in order to find errors. Then I can implement a ReturnCodeValidator if I want. If this is difficult to do in a general sense, maybe custodian can have some special implementation for checking for non-zero return codes. Although return codes are not standardized, I feel it is nevertheless a very strong convention that exit codes of nonzero indicate a problem.

shyamd:
I think we're trying to enable an exception rather than a normal use case. A return code for most unix programs means a serious problem occured that requires user intervention or the user intervened. 

The default behavior for custodian in these situations should be to throw an exception. For the rare cases where people need to handle return codes, e.g. @xhqu1981 , they have custom job classes and our solution should be made in such a way that we don't affect them too much. Those kinds of changes can be made as part of a major version change, not a bug fix.

computron:
I think it's a normal use case for custodian to run an application that throws a non-zero return code, and then use custodian to change some inputs and rerun (hopefully getting a zero return code this time).

But I also think it's a normal use case to want custodian to have a validator that checks the return code and that fails validation if the return code is not zero (independently of doing detailed examination of specific output files with String search, etc.)

xhqu1981:
QChem handler does check for return code. However, I don't think an addition of ReturnCodeValidator will jeopardize it.

computron:
@xhqu but in QChemHandler you are getting a return code by doing string search on a specific output file

I mean more from the process itself. For example, I could run the "cp" command with custodian (this certainly does not have output files), and if cp returned a non-zero exit code, I could mark that "cp" command as not having completed correctly.

xhqu1981:
I am sorry I confused you. I mean a ReturnCodeValidator would be fine with the current QChem handler. QChem handler is not something we need to worry about to add a ReturnCodeValidator.

computron:
@xhqu1981 I know you were not suggesting to add QChemHandler to ReturnCodeValidator. I am just saying that there is currently no validator, or the possibility of implementing a validator, that *directly* operates on the signal returned by a Process. Even a QChemHandler which you claim checks the return code, only does so by parsing an *output file* - it does not *directly* operate on the signal returned by the Process...

shyamd:
@computron 
@xhqu1981's QChem Job itself is different. It doesn't return a Popen object. He runs the code in the job's run method, deals with errors, etc, and then returns return code after Qchem is all said and done to the main custodian object.

xhqu1981:
Thanks for clarifying it. I think the return code design is better to be
considered as a general approach. It should be independent of QChem or VASP.

shyuep:
Communications between Custodian and Validator goes one way, i.e., Val -> Custodian. Custodian cannot pass return codes to the validator. 

There are a few ways we can implement this, but it would violate a lot of design principles. One is for Custodian to log return codes (or any other thing) to some global variable andd/or environmental varaible. Then the Validator can access that global variable and do something.

Alternatively, VaspJob can be modified to output a file to show that an error has occured.

In general, acting on return codes is usually not a good strategy. Fixes are based on outputs actually seen. I am not sure how let's say you get a return code = 63 on running VASP, but there is nothing visibly wrong with the outputs, that you are going to do anything to fix the run.

shyuep:
I just pushed a change to add an option to terminate on a non-zero return code. This defaults to **True**. If MP has a way to fix a non-zero return code, set it to False and implement a Validator that can check for errors to fix them.

shyuep:
I also added back the test that @shyamd wrote.

PR #47, opened 2017-02-19
Add terminate function (for SLURM) built in custodian to support atomate

specter119:
## Summary

* add terminate function to end job step (called by srun) for SLURM


shyuep:
I am unclear as to the reason for this functionality. Custodian does not assume any queuing behavior. The terminate_func is used to terminate actual programs like Vasp, matlab, etc. It shouldn't be used to call a specific cancellation for a queue job. You will notice there are no PBS equivalents of qdel. Why would then custodian need to implement scancel?

specter119:
@shyuep for the reason in https://github.com/hackingmaterials/atomate/pull/68, but I think I got a wrong idea to add the terminate function directly.

PR #56, opened 2017-07-06
Adds error handler for VASP when calculating electron localization function

mkhorton:
## Summary

Adds error handler when `LELF = True` is set in INCAR, which is incompatible with k-point parallelism and non-collinear calculations. Action is to set `LELF = False`, on the assumption that these other features take precedence (so that LELF can be enabled by default in high-throughput workflows).


shyuep:
Not sure I want to do this. LELF is a very specific tag that you enable only when you want the local electron potential. If someone sets it and you just turn it off to fix KPAR, the person would be confused why his LELF tag is being changed.

mkhorton:
We could invert it and set KPAR to 1 instead? Was thinking about using LELF by default in the same way that the AECAR files are generated by default by MPStaticSet in case someone wants to perform Bader analysis. No specific application in mind at present, but it seems like a cheap file to calculate and it's small too.

computron:
Maybe a separate handler instead of grouped with VaspErrorHandler would help?

shyuep:
I would rather switch KPAR to 1 than to turn LELF off. KPAR is a parameter that only affects speed, not success.

mkhorton:
That sounds sensible, I've revised the PR.

PR #58, opened 2017-07-28
Auto-continue and walltime handler refactor

montoyjh:
## Refactoring of continue functionality and walltime handler.

I've been using some of the wall-time and auto-continue functionality and thought it could use refactoring.  It might be useful to get some feedback on these before merging though.  The auto-continue functionality previously relied on the WalltimeHandler changing permissions in the STOPCAR so that VASP couldn't delete it, and then using the STOPCAR as a flag to trigger the copying of the CONTCAR to the POSCAR and the setting of ISTART to 1 in the INCAR.  Here's a summary of what I've changed.

**Continuation**
* VaspJob continuation functionality works by writing a file (continue.json) that stores a set of *actions* to be undertake by a VaspModder if custodian starts and detects that file in the working directory.  By default, these actions are moving the CONTCAR to the POSCAR and setting ISTART to 1 in the INCAR, but could be specified to be something else by giving the auto_continue kwarg a list of specified actions in the invocation of VaspJob.
* In postprocessing for VaspJob, continue.json is deleted if present so that future jobs run in the same directory don't take any of the prescribed actions.
* Note that this works independently of the walltime handler.  In principle, if custodian is stopped and isn't able to properly exit, the continue.json should persist and the job can be restarted using the prescribed actions.  The previous run is also backed up at the beginning of a continuation job.

**Walltime handling**
* This is refactored such that it monitors the OUTCAR for timings and gets them explicitly, rather than estimating from the OSZICAR.  This is necessary if the user is running multiple custodian jobs in the same batch job.
* Also refactored so that JOB_START_TIME can be set as an environment variable such that an accurate timing can be assessed if multiple custodian jobs are being run in the same batch job.  This can be achieved by using `export JOB_START_TIME=``date``` at the beginning of such a batch script.
* I removed the permissions changing procedure in the WalltimeHandler, since it no longer interacts directly with the auto-continue functionality.

Unittests have been refactored appropriately (and a few others with monty functionality for file handling without changing the static files).

## TODO:
* I'm not sure exactly what actions should be undertaken to do a NEB restart, so I didn't refactor VaspNebJob, but if given guidance, can do so in this PR.


mkhorton:
This looks really useful -- I was going to ask about how we're handling VaspDrone picking up the wrong POSCARs from continued jobs, but saw @shyamd's already taken care of this in https://github.com/hackingmaterials/atomate/pull/137 👍 

How does this tie in with FireWorks? Would a job that hits a walltime limit still be marked as fizzled and require user intervention to continue?

custodian/vasp/handlers.py
```diff
@@ -935,62 +934,46 @@ def __init__(self, wall_time=None, buffer_time=300,
...
             self.wall_time = None
         self.buffer_time = buffer_time
-        self.start_time = datetime.datetime.now()
+        if "JOB_START_TIME" in os.environ:
+            # Enables use of setting JOB_START_TIME as the start
+            # time to use for the walltime handler.  The easiest
+            # way to use this is by executing export JOB_START_TIME=`date`
+            # in the environment (e. g. in a queue submission script)
+            self.start_time = datetime.datetime.strptime(
+                os.environ["JOB_START_TIME"], "%a %b %d %H:%M:%S %Z %Y")
```
shyuep: I am not sure the necessity of this. The problem is that the actual time is measured by python. starttime is initialized whenever this class is init. I don't think there is a significant difference. If you throw in an externally measured time and for some reason, that time is not the same as the python measured time, then you have a mismatch.

custodian/vasp/handlers.py
```diff
@@ -935,62 +934,46 @@ def __init__(self, wall_time=None, buffer_time=300,
...
-                    # Intelligently determine time per ionic step.
-                    o = Oszicar("OSZICAR")
-                    nsteps = len(o.ionic_steps)
-                    time_per_step = total_secs / nsteps
-                except Exception:
-                    time_per_step = 0
+                # Determine max time per ionic step.
+                outcar.read_pattern({"timings": "LOOP\+.+real time(.+)"},
+                                postprocess=float)
+                time_per_step = np.max(outcar.data.get('timings')) or 0
```
shyuep: Do you have sufficient confidence in the OUTCAR reported timings? I have encountered situations in the past (long ago) where the reported times are not the same as the actual physical time.

custodian/vasp/jobs.py
```diff
@@ -162,20 +167,24 @@ def setup(self):
...
+                # Default functionality is to copy CONTCAR to POSCAR and set
+                # ISTART to 1 in the INCAR, but other actions can be specified
+                if self.auto_continue is True:
+                    actions = [{"file": "CONTCAR", 
+                                "action": {"_file_copy": {"dest": "POSCAR"}}},
+                               {"dict": "INCAR",
+                                "action": {"_set": {"ISTART": 1}}}]
+                else:
+                    actions = self.auto_continue
+                dumpfn({"actions": actions}, "continue.json")
```
shyuep: I don't have a strong view about this. In fact, I am not even sure what the auto_continue is used for.

custodian/vasp/handlers.py
montoyjh: They seem to be pretty accurate in my experience if you use "real time".  "CPU time" is all over the place in my experience, and I'm not really sure why.

custodian/vasp/handlers.py
montoyjh: The problem is if there are multiple custodian runs in the same batch job.  Let's say you run two custodian jobs or a lengthy preprocessing step and a custodian job in the same batch job that both take around 16 hours but you have a 24 hour walltime.  The second will initialize the start_time at the beginning of the *custodian job,* rather than using a time that represents when the batch job began, which was presumably sixteen hours before, and won't trigger the WalltimeHandler appropriately.

It's definitely true that there's a mismatch between when the batch job starts and the custodian job starts, but the point of this is that providing an environment variable from the batch environment is an appropriate reference for the batch walltime limit, rather the one in python that's relative to the custodian job.  It also doesn't change the default behavior (which is to initialize via python), just provides a user with a route to intervene in the aforementioned use case.

montoyjh:
In principle one could also refactor the VASP drone to make a calcs_reversed entry for each of the prev_calcs that are backed up, so you get basically the whole calculation, although this might be a bit wonky still with things like errors which sometimes do the same thing as the restart (i. e. copy CONTCAR to POSCAR).

As far as fireworks, the nice thing about this is that it should just depend on the presence of the files from the previous run that was cut short.  So task-level reruns with file copy should work fine.  The only thing that's necessary for it to pair more robustly there is a bit of expanded functionality for recovering lost runs (I'm in the middle of this now), but that's a fireworks issue.

custodian/vasp/handlers.py
montoyjh: also, ideally there would be a SLURM/PBS/etc. environment variable that one could use for this, but doesn't seem to be the case (at least for SLURM).

shyuep: OK. I would suggest this. Instead of having a wonky dependence on external timings, the code should set an environment variable CUSTODIAN_START_TIME. If this variable does not exist, it will be set to datetime.datetime.now(). Subsequent init of WalltimeHandler will use the os.environ version.

The reason is that I don't like that we have to specially invoke JOB_START_TIME in an external script.

shyuep: So basically,

```
if "CUSTODIAN_WALL_START_TIME" in os.environ:
    self.start_time = os.environ["CUSTODIAN_WALL_START_TIME"]
else:
    self.start_time = datetime.datetime.now()
    os.environ["CUSTODIAN_WALL_START_TIME"] = self.start_time
```

montoyjh: Yep, I've changed in the latest commit.

PR #60, opened 2017-10-12
Drift Error Handler

shyamd:
Error handler for excessive drift for VASP

* Uses some basic logic to slowly increase the augmentation grid to better calculate forces and enable less drift
1.) Adds ADDGRID which internally doubles the augmentation grid
2.) Sets PREC to High so that ENAUG controls the augmentation grid and then sets that to double ENCUT. ADDGRID is still TRUE so this should be a 4x the normal augmentation grid
3.) Continues to ramp up the augmentation grid by a user supplied factor

* Will get max_drift from EDIFFG if not set.
Default behavior is just set max_drift to be -1*EDIFFG, so this expects that force convergence is on. Basically useless if it's not. Does increasing the augmentation grid density do anything if energy convergence is used?


shyamd:
Any comment on this @shyuep ?

PR #69, opened 2018-01-10
Add FEFF submodule

czhengsci:
## Summary

Add submodule for FEFF calculation error correction with unittest


czhengsci:
The code quality review checks failed is because of the usage of subprocess module in running FEFF calculation script.

Issue #75, opened 2018-04-20
ALGO=All method for correcting unconverged electronic loops


computron:
## System

* Custodian version: master

## Summary

* If a job has not reached electronic converged (UnconvergedErrorHandler), the fix procedure in custodian is:

```
                new_settings = {"ISTART": 1,
                                "ALGO": "Normal",
                                "NELMDL": -6,
                                "BMIX": 0.001,
                                "AMIX_MAG": 0.8,
                                "BMIX_MAG": 0.001} 
```

Recently, I've had pretty good luck with just setting ```ALGO=All``` when a job is unconverged rather than making all these changes (This is in fact how UnconvergedHandler already deals with SCAN failures, but I think it's more general than that). If it were just up to my qualitative assessment, I would first try ```ALGO=All```; if that still doesn't work, to try making all the various changes above.

Downside: I don't have any hard facts to support my case.

Perhaps @shyamd or @montoyjh or @mkhorton or @tschaume can support or deny.

If we do want to switch to trying ALGO=All first, I'd be happy to do the implementation. So it's just a matter of deciding.

computron:
So maybe we want to keep normal as the first step, then simply add "All" as a new step if that one fails. The only issue is that currently custodian is also changing all these various mixing parameters which might mess things up (or might not, who knows)

shyuep:
The choice is ALGO is basically a competition between speed and reliability. As far as I know, no one has done a proper study of the difference in speed between Fast and Normal and All. Specifically, where does Normal fail?

shyamd:
I implement the ladder in case we starts off at something looser. I didn't want to jump from VF to All, for instance, but just incrementally trade off speed and stability until it worked. This might not be the most efficient.

mkhorton:
I'd caution against including VeryFast on the ladder for HT use, it exclusively uses RMM-DIIS which can fail in dangerous ways (it converges, but can converge to an incorrect result).

PR #76, opened 2018-04-20
Better convergence handling - VASP

shyamd:
Suggested method for laddering from VeryFast to All iteratively if electronic convergence is not met in VASP:
Solves this feature request - #75 
- Removes changes in mixing params
- Moves from VeryFast to Fast to Normal to All


shyamd:
So the handlers have different mixing parameter logic based on whether its the UnconvergedErrorHandler or NonConvergingErrorhandler:

I have no clue for which one to use since even the VASP manual is pretty vague about this.

computron:
The logic you've implemented in this PR (ladder down the Algo, then change mixing parameters only after hitting ALGO=All rather than up front) is in-line with what I would do based on my qualitative experience. But it is not really justified by real data. Up to @shyuep whether he wants to merge I guess, but I personally don't have any issues with it.

shyamd:
@shyuep,
Any chance you're going to accept this? Anecdotally, I've had a number of weird structures, particularly HOIP substitutions converge using Algo  = All under regular GGA+U that didn't with fast or normal, and Algo = ALL was slower than the other two per SCF, so I believe this laddering is, in fact, a speed/stability trade-off.

shyuep:
I need to see data on the tradeoffs. I cannot make a decision based on this alone. I would say that if HOIP requires different treatment, by all means have a different VASP handler for them. 

I am confused by what you mean by GGA+U for HOIP. As far as I know, there are no U elements in HOIP, unless you are referring to the literature proposed value of U=8eV for Pb (which frankly sounds ridiculous to me).

computron:
Any update on this PR? Curious whether it will be part of main custodian

shyuep:
My previous comment still stands.

shyamd:
Here's the results of a structured test on perturbed silicon supercell. I repeated this 4 times to get an average for the LOOP times and number of LOOP itterations for SCF.

- 432 Atom super cell (6x6x6)
- All positions pertubed by 0.05 A
- Static calculation to stress test the SCF convergence
- Max SC Steps set to 200
- VASP 5.4.4 with MPRelaxSet 
- 1 Cori Haswell Node, 24 total MPI tasks
- Max memory is total for all 24 tasks measured by SLURM


|           | GGA                     |              |                 | SCAN                    |              |                 |
|-----------|-------------------------|--------------|-----------------|-------------------------|--------------|-----------------|
| Algorithm | Average  LOOP Time  (s) | Num SC steps | Max Memory (GB) | Average  LOOP Time  (s) | Num SC steps | Max Memory (GB) |
| Very Fast | 72                      | 34           | 19              | 497                     | Failed       | 43              |
| Fast      | 120                     | 31           | 19              | 493                     | Failed       | 43              |
| Normal    | 168                     | 25           | 19              | 496                     | 104          | 43              |
| All       | 192                     | 22           | 23              | 776                     | 68           | 58              |

PR #78, opened 2018-06-05
QChem updates for Atomate

samblau:
## Summary

Small tweaks that were found to be necessary when building QChem functionality into Atomate and testing workflows.

* Standardized default input and output filenames. 
* Standardized variable name for scratch directory.
* Added option to ignore connectivity when flattening negative frequencies.

We would greatly appreciate if this could be merged as soon as possible and that a new version could be released immediately afterwards so that we can do a pull request for Atomate before Codebusters tomorrow. Thank you!

### DIFF
--- a/custodian/qchem/new_handlers.py
+++ b/custodian/qchem/new_handlers.py
@@ -29,8 +29,8 @@ class QChemErrorHandler(ErrorHandler):
     is_monitor = False
 
     def __init__(self,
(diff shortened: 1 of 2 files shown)


samblau:
All checks have passed. I believe this is ready to be merged in. We would greatly appreciate if it could be merged and a new version could be released this evening so that we're ready to go for Codebusters tomorrow morning. Thanks!

PR #81, opened 2018-08-08
CircleCI OpenBabel for QChem tests

samblau:
CircleCI now builds with python 3.6 and installs OpenBabel such that all QChem tests can be run


samblau:
All tests now pass. Note that I had to remove coveralls as it was throwing an error.

PR #129, opened 2019-10-30
Small change to chgcar validation

jmmshn:
The old way incorrectly flags some calculations as invalid when the AECCAR spikes too much in near the atomic cores.
Now only flags obviously broken charge densities and allows the flagging based on the difference to be tunable by the user.


shyuep:
Thanks.

PR #130, opened 2019-11-01
UnconvergedErrorHandler deprecated change_algo

acrutt:
change_algo input of UnconvergedErrorHandler is deprecated. Added a custom as_dict method as a fix to preserve backwards compatibility of Fireworks with older versions of Custodian that include change_algo.

@mkhorton was consulted for this fix. Note adding a logger warning was considered but this was not straightforward to incorporate.

### DIFF
--- a/custodian/vasp/handlers.py
+++ b/custodian/vasp/handlers.py
@@ -1055,6 +1055,17 @@ class NonConvergingErrorHandler(ErrorHandler):
         # Unfixable error. Just return None for actions.
         else:
             return {"errors": ["Non-converging job"], "actions": None}
+        
+    @classmethod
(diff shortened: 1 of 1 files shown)


mkhorton:
Thanks @acrutt! Yes, we have a lot of old FireWorks still to run which have `change_algo` present.

A broader fix could be to update monty itself to use `inspect` and only pass kwargs that are actually present in the initializer, but we can leave that for a later date, since we've encountered this bug in other contexts also.

mkhorton:
@acrutt two trivial pycodestyle issues to address: https://travis-ci.org/materialsproject/custodian/builds/606200763#L755

Issue #133, opened 2019-11-14
[VASP] Add handler for PSMAXN warnings and associated failures


rkingsbury:
## Summary

* VASP failure associated with PSMAXN warning is not recognized as an error by Custodian, triggering misleading ValidationError

## Details 

With certain combinations of ENCUT, LREAL, and pseudopotentials, VASP issues the warning

`WARNING: PSMAXN for non-local potential too small`

In some cases vasp still runs successfully, in other cases it will fail, e.g.

```
WARNING: PSMAXN for non-local potential too small
 LDA part: xc-table for Pade appr. of Perdew
 POSCAR, INCAR and KPOINTS ok, starting setup
 REAL_OPT: internal ERROR:         -32         -32         -32           0
 VASP aborting ...
 REAL_OPT: internal ERROR:         -32         -32         -32           0
 VASP aborting ...
 REAL_OPT: internal ERROR:         -32         -32         -32           0
 VASP aborting ...
 REAL_OPT: internal ERROR:         -32         -32         -32           0
...
```

It appears that Custodian does not recognize the above type of failure as an error. As a result, `_run_job()` will attempt to validate the output of the calculation (which never ran in the first place and therefore never generated output) and raise a `ValidationError` .

```
Traceback (most recent call last):
  File "/global/u2/r/rsking84/.conda/envs/cms/code/fireworks/fireworks/core/rocket.py", line 262, in run
    m_action = t.run_task(my_spec)
  File "/global/u2/r/rsking84/.conda/envs/cms/code/atomate/atomate/vasp/firetasks/run_calc.py", line 211, in run_task
    c.run()
  File "/global/u2/r/rsking84/.conda/envs/cms/code/custodian/custodian/custodian.py", line 378, in run
    self._run_job(job_n, job)
  File "/global/u2/r/rsking84/.conda/envs/cms/code/custodian/custodian/custodian.py", line 502, in _run_job
    raise ValidationError(s, True, v)
custodian.custodian.ValidationError: Validation failed: VasprunXMLValidator
```
The ValidationError is very difficult to troubleshoot without running vasp manually. In this situation, the contents of `vasp.out` are empty and `std_err.txt` contains only

`srun: fatal: Can not execute vasp_std`

## Suggested solution

Information about the PSMAXN warning and associated failures is scarce, but there appear to be several possible fixes:

1. Set LREAL=FALSE (expand the basis set in reciprocal space instead of real space)
[...]

PR #134, opened 2019-11-14
Handle PSMAXN / REALOPT errors

rkingsbury:
Add error handling for PSMAXN errors by setting LREAL=FALSE. See #133 .

Also note that several of the existing test files contain the PSMAXN warnings, so new tests are not required for this handler.


rkingsbury:
Updating the code to handle only the REALOPT error, not the PSMAXN warning. After further testing, it's evident that many calculations are able to complete with the PSMAXN warning, so we don't want Custodian to kick in and restart the calculation just because that warning is present.

rkingsbury:
Closing because REAL_OPT is already handled by the `real_optlay` handler

PR #138, opened 2019-11-20
[WIP] Update teterror handling for KSPACING

rkingsbury:
## Summary

VASP now [recommends](https://www.vasp.at/wiki/index.php/KSPACING) setting KSPACING in the INCAR rather than manually supplying KPOINTS files. This PR updates the teterror handler which is invoked when there are not enough kpoints to enable the tetrahedron integration scheme (ISMEAR=-5).

The solution is to lower the KSPACING value to 80% of the original setting (resulting in an approximate doubling of the no. of kpoints)

See also pymatgen PR # XXX


custodian/vasp/handlers.py
```diff
@@ -160,12 +161,21 @@ def correct(self):
         vi = VaspInput.from_directory(".")
 
         if self.errors.intersection(["tet", "dentet"]):
-            actions.append({"dict": "INCAR",
-                            "action": {"_set": {"ISMEAR": 0, "SIGMA": 0.05}}})
+            if incar.get("KSPACING",None) is not None:
```
mkhorton: This can be simply: `if incar.get("KSPACING"):`

mkhorton:
There seems to be more in this PR than just the KSPACING, e.g. metagga_opt additions, is this intended?

PR #149, opened 2019-12-17
Format project with black and update .gitignore

rkingsbury:
This is a housekeeping commit that updates .gitignore to ignore the mypy cache and also formats the entire project using the black formatter. Other team members and I are finding black to be a very useful time-saver now that our CI is enforcing stricter linting.

My intention with this change is to facilitate code review of future contributions. black does not support formatting selections of text, only entire files, so if I want to format new code with black, many other changes (unrelated to the new code) are generated. This can obscure what has actually changed.

If desired I can open similar PRs for pymatgen and fireworks.


rkingsbury:
Apparently black does not catch/correct pycodestyle "Whitespace before :" errors...

shyuep:
This is fine with me for now. But I think we need not apply to pymatgen first since that has already been completely reformatted. The remaining errors have to do with pydocstyle, which black will not fix.

PR #152, opened 2020-04-27
Validators and Jobs for Lobster

JaGeo:
Hi!

We (@gpetretto +I) are working on integrating Lobster workflows into atomate. Therefore, we would like to include new validators and jobs for Lobster into custodian.

We have also written tests. Please let me know if we should do any changes.

Best, 
JG


custodian/lobster/__init__.py
```diff
@@ -0,0 +1,12 @@
+from __future__ import unicode_literals
+
+"""
+This package implements various VASP Jobs and Error Handlers.
```
shyuep, line 4: Change this doc to say lobster?

custodian/lobster/handlers.py
```diff
@@ -0,0 +1,82 @@
+import os
```
shyuep, line 1: There needs to be a module doc.

custodian/lobster/jobs.py
```diff
@@ -0,0 +1,81 @@
+import logging
```
shyuep, line 1: Module doc.

test_files/lobster/vasp_lobster_output/FW_submit.script
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
```
shyuep, line 1: I don't think this file should be in here.

test_files/lobster/vasp_lobster_output/FW.json
```diff
@@ -0,0 +1,136 @@
+{
```
shyuep, line 1: Is this file even necessary. Custodian stands separate from Fireworks.

test_files/lobster/lobsterouts/lobsterout.crash
```diff
@@ -0,0 +1,42 @@
+LOBSTER v3.2.0 (g++ 5.4.0)
```
shyuep, line 1: Not sure what this file is. Or the .keep file.

test_files/lobster/vasp_lobster_output/FW_submit.script
JaGeo: will remove this.

test_files/lobster/vasp_lobster_output/FW.json
JaGeo: will remove this as well.

test_files/lobster/lobsterouts/lobsterout.crash
JaGeo: will remove the .crash file - the .keep was there to create the folder in the first place. I need the empty folder for the test.

custodian/lobster/__init__.py
JaGeo: okay

custodian/lobster/handlers.py
JaGeo: okay

custodian/lobster/jobs.py
JaGeo: okay

test_files/lobster/lobsterouts/lobsterout.crash
JaGeo: Do you have a suggestion how I can add an empty folder to git without the .keep file? Thanks!

shyuep: This is fine. But why do you need an empty folder for the test?

JaGeo: You are correct. I just need a folder without a "lobsterout" file to check that the validator is working. I will use one without lobsterout instead. Thanks.

JaGeo:
Is there anything else I should do? 

Best,
JG

shyuep:
Thanks.

JaGeo:
Thank you very much. I forgot to ask: Would it be possible to release a new custodian version?

shyuep:
Done.

PR #155, opened 2020-10-23
new handlers for SCAN workflow

rkingsbury:
## Summary

New handlers to support the new SCAN workflow

1. `LargeSigmaHandler` - when ISMEAR > 0, checks that the magnitude of the entropy is less than 1 meV/atom (per [VASP docs](https://www.vasp.at/wiki/index.php/ISMEAR])), and if not, reduces SIGMA. Applies to any calculation regardless of the functional.
2. `IncorrectSmearingHandler` - checks a calculation to ensure that metals (bandgap=0) were run with appropriate smearing settings, and reruns if needed. Applies to any calculation regardless of the functional. 
3. `ScanMetalHandler` - checks a calculation to ensure that metals (bandgap=0) were run with appropriate KSPACING parameter, and reruns if needed. Applies to SCAN calculations only.

## Notes

While `ScanMetalHandler` could in principle be applied to any calculation (not just SCAN), it applies KSPACING settings that are specific to `MPScanRelaxSet`. Hence, it may not be compatible with other calculation types.


rkingsbury:
@shyuep can you please let me know if you have any feedback on these new handlers?

shyuep:
I do not have a major issue with them.

PR #157, opened 2020-11-04
shorten err msg for vasp6. Closes #156.

### LINKED ISSUE #156: Some errors are not detected in VASP6
## System

* Custodian version: latest master branch as of 2020-11-01
* Python 3.8
* VASP 6.1.1

## Summary

`VaspErrorHandler` fails to detect certain error categories in VASP 6 that were detected properly in VASP 5. This appears to be a result of formatting changes associated with the error messages. My understanding is that custodian reads stdout line-by-line, but some of the error messages have been reformatted in such a way that they now break across multiple lines and hence are not caught.

See example below for the `point_group` error. I have also observed this for `inv_rot_mat` message, and I suspect it could affect other handlers as well.

## Error message

VASP6:

```
|     VERY BAD NEWS! internal error in subroutine IBZKPT: Error: point        |
|     group operation missing 2                                               |
```

VASP5:
```
 VERY BAD NEWS! internal error in subroutine IBZKPT:
 Error: point group operation missing       2
```

The `point_group` error looks for the entire message on a single line:
```
"point_group": ["Error: point group operation missing"],
```

## Suggested solution (if known)

It's not clear to me what the best solution is. Ideally we would move away from reading one line at a time and instead detect the entire string, regardless of where a line break occurs, but that may be difficult in practice. Alternatively, perhaps we need to shorten the error messages that are detected so that custodian will catch them in both VASP5 and VASP6.

rkingsbury:
Shorten error messages for 2 handlers so they are detected in VASP6. See #156.

### DIFF
--- a/custodian/vasp/handlers.py
+++ b/custodian/vasp/handlers.py
@@ -74,7 +74,7 @@ class VaspErrorHandler(ErrorHandler):
             "Tetrahedron method fails (number of k-points < 4)",
         ],
         "inv_rot_mat": [
-            "inverse of rotation matrix was not found (increase " "SYMPREC)"
+            "rotation matrix was not found (increase " "SYMPREC)"
         ],
         "brmix": ["BRMIX: very serious problems"],
         "subspacematrix": ["WARNING: Sub-Space-Matrix is not hermitian in " "DAV"],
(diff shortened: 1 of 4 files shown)


shyuep:
I do need a unittest for this...

rkingsbury:
> I do need a unittest for this...

@shyuep Since both of these handlers already have unit tests that pass, what did you have in mind? Do you mean a test that checks vasp5 vs vasp6 behavior?

shyuep:
I need an example VASP error file and show that the code detects it.

rkingsbury:
> > I do need a unittest for this...
> 
> @shyuep Since both of these handlers already have unit tests that pass, what did you have in mind? Do you mean a test that checks vasp5 vs vasp6 behavior?

Actually I'm mistaken, `inv_rot_matrix` does not currently have a unit test, only `point_group` does. I can add one with a VASP6 output file.

shyuep:
Thanks

PR #159, opened 2020-11-06
Update BRIONS error handling

ab5424:
Hello,

as far as I understand, the modification of POTIM should speed up convergence. Copying CONTCAR to POSCAR should help, in the same way as it is already used for ZBRENT errors.

Also, I added np.round(), because I saw floating point "errors" (i.e. POTIM = 0.9999999999999999 or so) in my INCAR. This should fix it.

Thanks for your feedback.


shyuep:
I am unclear what the change is. I thought the POTIM is already being fixed. Is this just for the CONTCAR -> POSCAR copy? Also, pls use the python round method rather than np.round. 

Finally, psl add unittests for the changes. Thanks.

ab5424:
The actual change I wanted to introduce is that CONTCAR is copied to POSCAR, as is the case with the ZBRENT handler. When I had convergence issues due to POTIM being too small, custodian started more than ten times from the original POSCAR, but when CONTCAR was copied into POSCAR, it resulted in fewer than 5 erroneous runs.

The round() is only a cosmetic fix. 

As for the test: Is there any test for the BRIONS handler yet? Unfortunately, I couldn't find it, I think it would just have to be extended (I have never written tests before, sorry for the confusion).

PR #161, opened 2020-11-30
detect POSMAP and BZINT, and SGRCON errors in Vasp 6

rkingsbury:
Small modification to VASP `handlers.py` in order to properly detect the POSMAP, SGRCON, and BZINTS errors in VASP 6, which formats the error messages differently than VASP 5.

I also updated the posmap error handler so that it 1) will not repeat the same action indefinitely and 2) tries both reducing and increasing the value of SYMPREC, as suggested by the VASP error message.


mkhorton:
Thanks for catching this; as previously, given the line breaks are different, is there a more general fix? I suspect many more handlers are affected too.

rkingsbury:
I think the two most viable solutions (that don't involve a lot of refactoring the guts of custodian) are 1) shorten the error messages to just the name of the error wherever possible (as I have here) or 2) list out multiple possible messages to catch for each error as currently done for `zbrent`:

```
"zbrent": ["ZBRENT: fatal internal in", "ZBRENT: fatal error in bracketing"],
```

I didn't realize option 2) was possible until this morning actually :)

rkingsbury:
@shyuep friendly reminder to please review at your earliest convenience. Thanks.

rkingsbury:
@shyuep I've just pulled the latest master so that the CI tests run properly on this PR (and all pass). Can you please merge at your earliest convenience? Also, I know you just did a release, but it would be helpful to do another one once this and #162 are merged so that I can finalize the associated atomate PR. Thanks!

PR #162, opened 2020-11-30
tweak LargeSigmaHandler correction

rkingsbury:
Tweak the correction made by `LargeSigmaHandler` so that it traverses the feasible range of SIGMA values in fewer steps, and ensure that SIGMA always remains at least 0.02. 

After running many calculations with `LargeSigmaHandler`, I found that this handler was often triggering multiple times on the same calculation, and in many cases causing `MaxCorrectionsPerJob` failures. This PR changes the handler so that a maximum of 3 corrections will be made (with larger adjustments at each step), leaving some cushion for other handlers to make corrections without hitting the default limit of 5.


rkingsbury:
@shyuep friendly reminder to please review at your earliest convenience. Thanks.

rkingsbury:
@shyuep I've just pulled the latest master so that the CI tests run properly on this PR (and all pass). Can you please merge at your earliest convenience? Also, I know you just did a release, but it would be helpful to do another one once this and #161 are merged so that I can finalize the associated atomate PR. Thanks!

PR #164, opened 2021-01-22
Updated the check method in the IncorrectSmearingHandler class

MichaelWolloch:
of `custodian.vasp.handlers` to allow static calculations with `ISMEAR = -5` for metals.

A detailed discussion can be found [here](https://matsci.org/t/custodian-incorrectsmearinghandler-should-allow-for-total-energy-calculations-for-metals-with-ismear-5/35157).

## Summary

Without this fix custodian does not allow static calculations for metals with the tetrahedron method. However, this is the recommended way to perform accurate total energy calculations in VASP for all materials.

The fix changes:
* Static calculations for metals with ISMEAR = -5 are now possible.
* The docstring of the class was updated accordingly.
* The fix simply checks if the NSW parameter (number of ionic steps) in the INCAR is <=1 and only triggers the error handling if it is.
* In contrast to the old version ISMEAR = -1 (Fermi smearing), -2 (partial occupancies are read in from the WAVECAR or INCAR file, and kept fixed throughout run), and -3 (perform a loop over smearing-parameters supplied in the INCAR file.) do not trigger the error handler any more. Only ISMEAR = -4, -5 (tetrahedron method without/with Blöchl corrections) trigger the correction.


shyuep:
Thanks. Can you provide a unittest pls?

MichaelWolloch:
Hej shyuep,
I have never written a unittest (this is also my first pull request). Can you point me to a similar example from custodian and tell me what I should do? 
Do you need a list of vasp files for which this handler should trigger?

rkingsbury:
Hi @MichaelWolloch , thanks for opening your first PR! Contributions from users like yourself are tremendously helpful for lessening the burden on our development team. Let me walk you through what we're looking for in the unit test:

1. Duplicate the `IncorrectSmearingHandlerTest` class in `/custodian/vasp/tests/test_handlers.py`. You could rename it something like `IncorrectSmearingHandlerStaticTest`
2. Duplicate the directory `test_files/scan_metal`. This contains a simulated INCAR, POSCAR, etc. that the handler will check. You can rename the duplicate directory something like "static_smearing"
3. Change the INCAR in that directory to have NSW=0 or NSW=1
3. point the `setUp` method of your class from step 1 to the dupliated directory
4. In the method called `test_check_correct_scan_metal` (in your new class from step 1), you just need to instantiate the class and then `self.assertFalse(h.check())`

To verify that this test works properly, you can `pip install pytest` on your local computer, then navigate to the custodian directory and run `pytest`.

Ideally, repeat the above procedure to create one more unit test, but this time set ISMEAR=-1, -2, or -3 in the INCAR (leave NSW=99), and verify that the handler is not triggered.

Finally, before we can merge your PR it will need to pass a series of linting (formatting) checks. Specifically, we lint with pycodestyle, flake8, mypy, pydocstyle, and pylint. You can install these on your local machine and run the checks before you push, or you can inspect the output of the CI tests above to see where the linting fails.

I hope this all makes sense. Thanks again for the effort on this and feel free to send questions.

custodian/vasp/handlers.py
```diff
@@ -1013,8 +1013,11 @@ def check(self):
         """
         try:
             v = Vasprun(self.output_filename)
-            # check whether bandgap is zero and tetrahedron smearing was used
-            if v.eigenvalue_band_properties[0] == 0 and v.incar.get("ISMEAR", 1) < 0:
+            # check whether bandgap is zero, tetrahedron smearing was used
+            # and relaxation is performed.
+            if (v.eigenvalue_band_properties[0] == 0 and
+                v.incar.get("ISMEAR", 1) < -3 and 
+                v.incar.get("NSW", 0) <= 1):
```
rkingsbury, line 1020: Looking at this, I realize that the logic is actually inverted. That's why the existing smearing test is failing. This should be `and v.incar.get("NSW", 0) > 1:` That way the handler will activate `check=True` for relaxations and not activate for statics.

MichaelWolloch: @rkingsbury, I am pretty emberrest by this mistake. Thanks a lot for catching this and also thanks for explaining how to do unit tests and linting correctly for this project.

I am guessing since the tests now pass with the updated logic and the correct pycodestyle, that I do not have to write a new `IncorrectSmearingHandlerStaticTest`? The INCAR in `test_files/scan_metal` has `NSW = 99` and `ISMEAR = -5`, and thus correctly triggers the handler I think.

Of course I am happy to follow your instructions for the new test if it is needed, even if the old one now completes OK.

rkingsbury: Thanks @MichaelWolloch; no problem. You're correct that after changing the logic the existing test now passes, and you've resolved all the linting errors which is good. 

It would still be preferable if you could write an additional unit test to confirm that `IncorrectSmearingHandler` is not triggered during static calculations. Would you mind giving that a try? I know writing tests is tedious, but it's critical to keeping our codes from breaking as they evolve.

MichaelWolloch: I will certainly do that. Hopefully I will find time today, otherwise you can expect something tomorrow! I am glad to be able to help a little, since I am profiting a lot from the work you all are doing within the materialsproject.

MichaelWolloch:
I now pushed 2 more unit tests to check if static calculations and Fermi smearing (`ISMEAR = -1`) pass. I followed @rkingsbury's guidance and the tests passed.

Initially I only modified the INCARs in the new directories `test_files/static_smearing` and` test_files/fermi_smearing`, and the tests failed. But than I realized that the handler reads the values from the vasprun.xml file and not the INCAR. I guess for a "real" test I should have rerun the vasp calculations with the new INCAR, but I just modified the vasprun.xml file instead. If that is a problem, please tell me and I will re-run the vasp calculations to produce a "proper" vasprun.xml.

rkingsbury:
Thanks @MichaelWolloch ! These tests look good to me. I'll defer to @shyuep on your question about the vasprun.xml, but if you just manually changed the INCAR settings (in the vasprun.xml) then I don't see a need to rerun the calculation.

MichaelWolloch:
> if you just manually changed the INCAR settings (in the vasprun.xml)

This is exactly what I did.

rkingsbury:
@shyuep can you please review this PR at your earliest convenience, and let us know if you have other changes to request?

MichaelWolloch:
Hi, I do not want to be a pest, but it has been two weeks now since I made the last changes. I am happy to do more (e.g. run new calculations for the unit tests). Maybe @shyuep can get involved again?
Cheers, Michael

shyuep:
Sorry for the delay. It has been merged.

MichaelWolloch:
No Problem, thanks a lot for considering my input and helping me to make it presentable.

PR #166, opened 2021-03-01
More gracefully catch a missing Openbabel installation

samblau:
## Summary

OpenBabel is a requirement for core functionality of QCJob and QChemErrorHandler, but it cannot be enforced as a true requirement because it cannot be installed via pip and instead must be installed with conda. However, at present, if OpenBabel is not installed, the resulting errors are difficult to interpret. This PR adds in just a few lines of code to catch a missing OpenBabel installation and exit with much a much clearer error message so the user knows that they need to install OpenBabel to proceed.


shyuep:
I am not sure why the output file has anything to do with an openbabel installation. Why not simply check for an openbabel install?

samblau:
That is a very fair question, and that's actually the first way I tried to implement it. However, if I check for an openbabel installation by doing a try/except with an import, then when I successfully import it, Flake8 complains that I never use my imported openbabel module. I'm open to suggestions of ways around this. Thoughts?

samblau:
A bit more context - the reason that we need openbabel to be installed in Custodian is because QCJob and QChemErrorHandler both depend on the presence of the "structure_change" key in the parsed output file, and that key will only be populated in pymatgen's QCOutput if openbabel is installed.

shyuep:
There are specific ways around it. https://flake8.pycqa.org/en/3.1.1/user/ignoring-errors.html

samblau:
@shyuep I believe this is not ready to merge

PR #169, opened 2021-05-19
QChem: handle basis_not_supported error

rkingsbury:
Adds handler for basis set not supported error to QChem. This error cannot be automatically corrected, so it is handled similarly to the existing `input_file_error`. No corrective action is taken.

See pymatgen #[2147](https://github.com/materialsproject/pymatgen/pull/2147). There should be no harm merging this PR before or after the pymatgen PR, but the test will not pass and obviously this PR will have no effect until #2147 is merged and the requirements.txt for custodian bumped to pick up the change.

### DIFF
--- a/custodian/qchem/handlers.py
+++ b/custodian/qchem/handlers.py
@@ -10,7 +10,7 @@ from pymatgen.io.qchem.outputs import QCOutput
 from custodian.custodian import ErrorHandler
 from custodian.utils import backup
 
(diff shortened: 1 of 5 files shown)


rkingsbury:
I guess tests will not pass until pymatgen #2147 is incorporated into a release and the CI here is updated

shyuep:
This has merge conflicts.

rkingsbury:
Thanks @shyuep; just resolved conflicts that I think were introduced by merging #170

Issue #92, opened 2019-03-21
Better EDWAV error fix


dkitch:
The current fix for the EDWAV error is to change the smearing scheme to ISMEAR = 0, which is not a very reliable fix. The origin of this error appears to be a bug in VASP when it is compiled with the default -O2 level of optimization in some versions of the intel compiler. Changing the optimization level to -O1 appears to reliably fix the issue. 

I suggest adding a warning or optional switch to an O1-version of VASP to fix the bug, instead of changing the smearing. For what its worth, EDWAV errors only show up when running ALGO = A or ALGO = D, which can't be run with the common ISMEAR = -5 scheme anyways.

shyuep:
Sure, by all means implement in custodian as a unfixable error.

dkitch:
I'm mostly suggesting this as documentation because this error has been really annoying for me for over 3 years of projects and there is very little documentation on it available. Recompiling VASP with a different optimization flag is not hard, and adding a warning message suggesting that that's the solution could save people a lot of frustration.

If you don't think that's in line with the purpose of custodian, then please close the issue. Otherwise, I can create a pull request at some point with the warning message implementation.

shyuep:
I agree this is in line with the purposes of custodian and I am in fact agreeing with your suggestion. What I am asking is that you change the EDDAV fix in custodian to be an "unfixable" error, which means VASP will immediately terminate. The error message can indeed be what you suggested, i.e., please recompile VASP with the O1 flag. I think this is preferable to custodian just blindly changing the ISMEAR and moving on, even with a warning. My experience is that once people are done with a calculation, they seldom go back and read warning messages. Hard termination is a better way.

Just a note that I personally think we (meaning me and associates) have been a bit too cavalier about ISMEAR settings. There are probably better ways to handle this. But that is a separate topic.

mkhorton:
Agreed we need to be more careful with ISMEAR and should probably look into that more carefully. Slightly unrelated, but I'm not sure the effect of LREAL Auto has been fully explored in the MP dataset either.

arosen93:
I agree with what's written here. I have encountered this issue countless times, and it's extremely compilation-dependent. It commonly crops up when Algo = All is used (also Algo = Damped, as mentioned by @dkitch). I don't think ISMEAR is the appropriate fix here. Instead, I think the better change to make is to switch from Algo = All to Normal _if_ a meta-GGA/hybrid functional isn't used. I've opened a PR to address this.

[...]

PR #177, opened 2021-11-14
[WIP] Better handling of large, flexible systems and meta-GGAs/hybrids

arosen93:
# Background
I have opened a PR for a number of additions to the VASP error handlers in Custodian. These new INCAR switches are based on my personal experience carrying out tens of thousands of DFT relaxations for metal-organic frameworks (https://materialsproject.org/mofs), albeit not via the use of Custodian. In most cases, I have tried to ensure that no changes are made to existing INCAR swaps. Rather, if the existing INCAR swaps do not resolve the issue, I have added new backup options.

I have labeled this WIP for a few reasons: 1) I need to write unittests for this, but I figured I'd see if there's any interest before I do a deep dive on the testing. 2) I'd like to test this out for a bit in production calculations to make sure nothing unusual crops up. Let's consider this the start of a conversation.

Below, I outline the proposed changes. I can also split this into multiple PRs if preferred.

# Main Changes
- zbrent: This error is perhaps the most important one I have addressed. Currently, Custodian switches to IBRION = 1 and copies CONTCAR to POSCAR. For materials that are large and have flexible structures, this is rarely sufficient. There is still too much numerical noise in the forces when this error occurs near the local minimum in the PES. A fallback option of tightening EDIFF to 1e-6 (if it was larger than this value) and increasing NELMIN to 6 is applied. These ensure that the forces are of higher quality. I have also created a new `vtst_enabled` variable, which gives the user an option to support VTST-related INCAR swaps if they have a VASP build with VTST. I've set `vtst_enabled = False` as the default for compatibility purposes. However, if `vtst_enabled` is set to `True`, the optimizer will switch to FIRE. In my experience, FIRE with the modified EDIFF and NELMIN has worked every single time to fix the ZBRENT issue, and I have received it hundreds of times.

- grad_not_orth: Depending on how you've compiled VASP, the Algo = All SCF convergence algorithm can sometimes lead to grad_not_orth errors. It's extremely machine and compilation-dependent. Switching to any other major SCF convergence algorithm resolves the issue in these cases. So, I've added a swap from All to Fast if this error comes up (unless dealing with meta-GGAs/hybrids, as discussed below). Related to #92. Also, I don't think changing ISMEAR is the way to go here.

- posmap: There was no `self.error_count["posmap"] += 1` line despite there being an if/elif switch based on the error count. I have added that in now. I'm not sure the second loop was ever reached?

- real_optlay: Currently, this INCAR swap implicitly assumes the user has set LREAL = Auto, but there is no guarantee this is the case. The LREAL = Auto --> True [for large systems] --> False chain now only takes place if LREAL != False in the first place.

Additionally, a few important changes were made to the SCF ladder in `NonConvergingErrorHandler`:
- If a meta-GGA functional is used, go straight to Algo = All. This is recommended in the VASP manual (https://www.vasp.at/wiki/index.php/METAGGA) and some of the meta-GGAs (e.g. the Minnesota functionals) explicitly say to set Algo = All for proper convergence. The `MPScanRelaxSet` already uses Algo = All as the default, but other users may not be so careful.
- If a hybrid functional is used, *do not* set Algo = Fast or VeryFast. Hybrid calculations do not support these algorithms even though no warning is printed (https://www.vasp.at/wiki/index.php/LHFCALC). Instead, do Algo = All if it's not already set. `MPHSERelaxSet` already uses Algo = All. In principle, if Algo = All doesn't work, one could switch to Algo = Damped with TIME = 0.5 per the VASP manual, but given the high cost of hybrids, I do not know that this is advisable in general and so I have left it as a comment.


arosen93:
This got a little larger than planned. I'm going to split this into multiple issues/PRs.

PR #180, opened 2021-11-15
[VASP] More appropriate grad_not_orth fix and new algo_tet error handler

arosen93:
1. Depending on how you've compiled VASP, the Algo = All (or Damped) SCF convergence algorithm can sometimes lead to `grad_not_orth` errors. It's extremely machine and compilation-dependent. Switching to any other major SCF convergence algorithm resolves the issue in these cases. So, I've added a swap from All/Damped to Normal if this error comes up (unless dealing with meta-GGAs/hybrids, in which case the Algo should probably not be swapped). Since it's largely a matter of how VASP was compiled and specifically occurs with Algo = All/Damped, it's highly unlikely that changing ISMEAR will fix the issue, so I have replaced it with the aforementioned Algo swap. See additional discussion in #92.

2. There is a related warning, however, when Algo = All is set and ISMEAR < 0 (see below) that typically leads to a VASP crash. While it often coincides with `grad_not_orth`, they are not the same errors. It is also distinct from the `tet` error, which occurs when the number of k-points is too low for the tetrahedron method. As such, I've introduced a new error message called `algo_tet` that captures this. It uses the old `grad_not_orth` fix of switching ISMEAR unless it looks like you're running a DOS calculation, in which case it doesn't touch ISMEAR.

```
|      ALGO=A and IALGO=5X tend to fail with the tetrahedron method           |
|      (e.g. Bloechls method ISMEAR=-5 is not variational)                    |
|      please switch to IMSEAR=0-n, except for DOS calculations               |
|      For DOS calculations use IALGO=53 after preconverging with ISMEAR>=0   |
|          I HOPE YOU KNOW, WHAT YOU ARE  DOING                               |
```


arosen93:
This one is ready to go. Tests are now included.

PR #182, opened 2021-11-15
[VASP] Update real_optlay logic to avoid LREAL = True

### LINKED ISSUE #181: real_optlay assumes LREAL = Auto when this is not necessarily the case
Currently, the `real_optlay` INCAR swap implicitly assumes the user has set LREAL = Auto, but there is no guarantee this is the case.

arosen93:
Closes #181:

`real_optlay`: For large systems, LREAL is changed to True first before changing it to LREAL = False in the current version of Custodian. There are two concerns with this logic. 

1. It was never checked that LREAL != False in the first place. We don't want LREAL = False to accidentally be changed to True as a "fix".

2. More importantly, the [VASP manual](https://www.vasp.at/wiki/index.php/LREAL) advises against using LREAL = True. The small speedup is likely not worth the hit to accuracy.

From the VASP manual:
> For LREAL=.TRUE., the real space projectors that have been generated by the pseudopotential generation code are used. This requires no user interference but is potentially very inaccurate.

> LREAL=.TRUE. | projection done in real space, use projectors on file (not recommended)

> We also strongly recommend to use only LREAL=Auto [for real-space projection]

With these two factors in mind, I have simplified the `real_optlay` check to just change LREAL to False if it wasn't already. This does make it so the `natoms_large_cell` kwarg is not used for anything, so we may want to consider removing that kwarg in the future.


arosen93:
This one is ready to go (tests are updated), once we decide what to do with the now unused `natons_large_cell` argument. I don't want to remove it in case it is used in other codes, but it's also not doing anything.

PR #183, opened 2021-11-15
[VASP] Increase posmap error count

### LINKED ISSUE #186: [VASP] [Bug] Second posmap error does not change SYMPREC
When the VASP `"posmap"` error occurs twice, [SYMPREC should be changed to 1e-4](https://github.com/materialsproject/custodian/blob/master/custodian/vasp/handlers.py#L413). However, it stays 1e-6 because the error counter was never updated. A test was also not written for the second SYMPREC error handler. 

Edit: Will be closed when #183 is merged.

arosen93:
Closes #186:

- Previously, `self.error_count["posmap"]` was never updated even though there were two sequential INCAR swaps to consider. This caused the second fix to never take place, and there was no test to check for it. There is now a test and the issue is fixed.
- I more clearly set `self.error_count["subspacematrix"] == 1` in the if/else block instead of relying on `else`.

### DIFF
--- a/custodian/vasp/handlers.py
+++ b/custodian/vasp/handlers.py
@@ -1081,7 +1081,7 @@ class LargeSigmaHandler(ErrorHandler):
 class MaxForceErrorHandler(ErrorHandler):
     """
     Checks that the desired force convergence has been achieved. Otherwise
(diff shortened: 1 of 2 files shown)


arosen93:
This one is done. Tests are updated.

PR #184, opened 2021-11-15
[VASP] More robust zbrent fix

arosen93:
`zbrent`: Currently, Custodian switches to IBRION = 1 and copies CONTCAR to POSCAR. For materials that are large and have flexible structures, this is rarely sufficient.* I do not change the original approach here, but rather I have added a _second_ fallback option of tightening EDIFF (per the recommendation in the stdout when this error occurs, shown below) and increasing NELMIN to 6 (i.e. between 4 and 8 as noted in the [VASP manual](https://www.vasp.at/wiki/index.php/IBRION#IBRION.3D1:_ionic_relaxation_.28RMM-DIIS.29.)) in this PR. These ensure that the energy/forces are of higher quality. See the two quotes from the stdout and VASP manual below for context:

```
 curvature:   0.00 expect dE= 0.000E+00 dE for cont linesearch  0.000E+00
 ZBRENT: fatal error in bracketing
     please rerun with smaller EDIFF, or copy CONTCAR
     to POSCAR and continue 
```

> Since the algorithm [IBRION=1] builds up an approximation of the Hessian matrix it requires very accurate forces, otherwise it will fail to converge. An efficient way to achieve this is to set NELMIN to a value between 4 and 8 (for simple bulk materials 4 is usually adequate, whereas 8 might be required for complex surfaces where the charge density converges very slowly). This forces a minimum of 4 to 8 electronic steps between each ionic step, and guarantees that the forces are well converged at each step.

If the new `vtst_fixes` argument is set to `True` (default: `False`), it will also switch to the [FIRE](https://journals.aps.org/prl/abstract/10.1103/PhysRevLett.97.170201) optimizer when correcting a second failed `zbrent` run. I have run into this error hundreds of times and had a near-100% success rate with FIRE, but it's not enabled by default for compatibility purposes and in case others disagree with this decision for some reason.

Custodian didn't have any tests at all for `zbrent`, so I have added those too.

*This is based on personal experience with thousands of calculations on porous materials (a large portion of structures at https://materialsproject.org/mofs run into this issue and aren't fixed by a simple CONTCAR --> POSCAR swap). There is still too much numerical noise in the energy when this error occurs near the local minimum in the PES.


arosen93:
This one is done. Added missing "zbrent" tests that were never included.

PR #197, opened 2022-01-13
New NBANDS not sufficient handler in VASP

### LINKED ISSUE #196: NBANDS VASP error without a handler
In VASP, there is an error shown below when NBANDS is set to a terribly odd value. As you can see below, NBANDS is ridiculous, and the calculation fails (I got here because I accidentally had MAGMOM = 2*nan in my INCAR and apparently that did some weird things with NBANDS). 

We should let VASP crash in this case because, indeed, something is terribly wrong. But we should still have a handler and print a warning to check the INCAR. I will add one shortly.

```
|     The number of bands is not sufficient to hold all electrons. I am       |
|     sorry, but this calculation does not make sense. If you want to         |
|     calculate only selected states, you need to set either EFERMI oder      |
|     EREF. EREF specifies around which energy states are supposed to be      |
|     calculated.                                                             |
|     I found NBANDS = -1073741632, NELECT = 250.                             |
```
[OUTCAR.txt](https://github.com/materialsproject/custodian/files/7865334/OUTCAR.txt)

arosen93:
Closes #196. 

## VASP Error
There is an unflagged error that reads: `The number of bands is not sufficient to hold all electrons`. This occurs if NBANDS is less than the number of electrons. Note that this is a different error than the `too_few_band` error, which occurs when NBANDS is physical but should nonetheless be increased. 

## Solution
I've made a new error handler called `nbands_not_sufficient` that simply logs the error. I've made it "unfixable" because it usually means something is terribly wrong with how the user set up their INCAR. We probably don't want to just naively increase NBANDS here because other things might be strange.

For context, this happened to me when I accidentally had "nan" in the MAGMOM line, which for some reason caused VASP to set NBANDS to a negative value, triggering the below error. Very bizarre.

## The VASP error message
```
|     The number of bands is not sufficient to hold all electrons. I am       |
|     sorry, but this calculation does not make sense. If you want to         |
|     calculate only selected states, you need to set either EFERMI oder      |
|     EREF. EREF specifies around which energy states are supposed to be      |
|     calculated.                                                             |
|     I found NBANDS = -1073741632, NELECT = 250.                             |
```


shyuep:
This can't be merged due to conflicts.

arosen93:
@shyuep: Thanks for the head's up. I figured that was possible. Should be fixed now.

PR #198, opened 2022-01-13
Better handling of VASP algo_tet error

### LINKED ISSUE #195: VASP: Update INCAR when algo_tet error occurs and Algo = All
Currently, when the `algo_tet` error occurs (i.e. you're using ALGO = All with ISMEAR = -4/-5), no INCAR swaps are made if a DOS calculation is being run. I did this out of caution, but I think it'd be better to implement an INCAR swap.
https://github.com/materialsproject/custodian/blob/e84d9d12377752fd180a90a5f98192bf500b24c6/custodian/vasp/handlers.py#L429-L436

A better solution would be to either:
- Change ALGO to Damped. The downside is that VASP specifically states to do this *after* a run with ISMEAR >= 0.
- Change ISMEAR to 0 and SIGMA to 0.05 anyway. The downside is that ISMEAR = -5 is often desired for a DOS calculation. This is what Custodian had done in the past before I split up `grad_not_orth` and `algo_tet` into two errors.

I am benchmarking both approaches and will implement a fix.

arosen93:
Closes #195.

## Issue
When ISMEAR = -5 and ALGO = All, the `algo_tet` error generally occurs since these settings are incompatible. During a DOS calculation with this error, the handler currently does nothing. I had set this up intentionally, but after running a lot of calculations, I've noticed that it can lead to an undesirable number of jobs failing if SCF convergence is challenging and ALGO gets moved from Fast --> Normal --> All frequently.

## Solution
This PR changes ISMEAR to 0 and SIGMA to 0.05, which was the default behavior a few versions back when this error was instead handled under the `grad_not_orth` handler (`grad_not_orth` can occur without `algo_tet` when there are compilation flag issues, and in that case changing ISMEAR does not fix the issue, which is why I split them up in #180).

--------
A better solution for `algo_tet` DOS runs would be to: 1) Set ISMEAR = 0, SIGMA = 0.05, and LWAVE = True and run VASP; 2) From the prior run (if successful), set ISMEAR back to -5, set SIGMA back to the old value, and set ALGO = Damped [this is one of the options recommended in the VASP error message]. However, this two-step process is substantially more difficult to implement in Custodian. The user can always rerun their DOS calculations if desired on their own.

### DIFF
--- a/custodian/vasp/handlers.py
+++ b/custodian/vasp/handlers.py
@@ -434,7 +434,8 @@ class VaspErrorHandler(ErrorHandler):
                 if vi["INCAR"].get("NEDOS") or vi["INCAR"].get("EMIN") or vi["INCAR"].get("EMAX"):
                     warnings.warn(
                         "This looks like a DOS run. You may want to follow-up this job with ALGO = Damped"
-                        " and ISMEAR = -5, using the wavefunction from the current job."
+                        " and ISMEAR = -5, using the wavefunction from the current job.",
+                        UserWarning,
                     )
                 actions.append({"dict": "INCAR", "action": {"_set": {"ISMEAR": 0, "SIGMA": 0.05}}})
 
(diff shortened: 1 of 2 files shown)


arosen93:
Going to merge this with #201 to prevent merge conflicts.

PR #201, opened 2022-01-16
Use ALGO = Fast when grad_not_orth occurs; have a backup for hybrid SCF convergence issues

arosen93:
- The `grad_not_orth` handler now checks for IALGO = 53-58 (in addition to ALGO = All/Damped), per the VASP error message.
- When I wrote the `grad_not_orth` handler, I had the INCAR swap be a change from ALGO = All/Damped to Normal. In this PR, ALGO is now changed to Fast instead. In hindsight, there was no particular reason to use Normal. The key is to just switch away from All/Damped, and Fast is usually -- well -- a faster option.
- In accordance with the VASP manual on [LHFCALC](https://www.vasp.at/wiki/index.php/LHFCALC), when a hybrid calculation has SCF convergence issues with ALGO = All, ALGO = Damped will be tried with TIME = 0.5. A test was added.
- I removed some extraneous test lines I had introduced with the `grad_not_orth` handler.


arosen93:
Going to combine this with #198 otherwise it's going to be a merge conflict disaster.

PR #210, opened 2022-03-16
Cp2k support

nwinner:
## Summary

This PR is to merge my personal custodian modules for running CP2K into the main custodian branch. This was originally a "personal" branch with lots of little commits. To keep version history clean, I can create a new branch for the PR.

* CP2K support following the vasp module's outline: 
* **jobs.py**: defines a standard cp2k job, including the ability to setup a job with modifications (used by atomate env_chk). Some constructor methods exist for defining double jobs (see below).
* **handlers.py**: The meat of the PR, defines a number of handlers for detecting and correcting CP2K job errors. Fairly intricate, and very distinct from those used by vasp.
* **validators.py**: Defines a base validator in case of new validators to be added. Also defines a standard validator that checks to see if a job has completed.
* **interpreter.py**: similar to that in the vasp module including an overwrite of the _modify() helper function for performing in place modifications of input objects.
* **utils.py**: useful utils mainly for the handlers.

## TODO

* Works with the pymatgen module for CP2K. Pymatgen pre-production CP2K module is in the main branch, but this uses a couple features from the production version, which are being PR'd at the moment.
* jobs.py: the constructors may be largely unecessary. That sort of functionality may be better off as leaving to atomate. Need to decide before production.
* utils.py: Some of the utils are no longer used, but I saw some utility in them in the past. If they are no longer used they can be deleted.
* Handler's are never perfect, but there may be some room for refinement with the ones that are listed here.


nwinner:
@shyuep Could you take a look at this and let me know if you'd like anything adjusted?

PR #220, opened 2022-05-17
BUGFIX: VasHandler: too_few_bands

yury-lysogorskiy:
## Summary

* check if NBANDS actually increased, increment it by one if int(1.1 * nbands)==nbands to avoid infinite try-and-error loop
Example: NBANDS=7 and  too_few_bands error encountered, then correction would be  int(1.1*nbands) = 7 again, so there is no update actually

Solution: 
```python
new_nbands = int(1.1 * nbands)
if new_nbands == nbands:
    new_nbands += 1
actions.append({"dict": "INCAR", "action": {"_set": {"NBANDS": new_nbands}}})
```


custodian/vasp/handlers.py
```diff
@@ -439,7 +439,10 @@ def correct(self):
                                 break
                             except (IndexError, ValueError):
                                 pass
-            actions.append({"dict": "INCAR", "action": {"_set": {"NBANDS": int(1.1 * nbands)}}})
+            new_nbands = int(1.1 * nbands)
+            if new_nbands == nbands:
+                new_nbands += 1
```
shyuep, line 444: I think a better way to write this is `new_bands = max(int(1.1*nbands), nbands+1)`? It is one line instead of three.

shyuep:
It's ok. I will fix it on my end. I will merge this now.

custodian/vasp/handlers.py
yury-lysogorskiy: yep, that is even better

PR #223, opened 2022-06-05
Skip Q-Chem tests if `openbabel` is not installed and not running in CI

janosh:
`openbabel` is often a hassle to install, especially outside a `conda` env. This PR changes the tests under `custodian/qchem/tests` to only run if `openbabel` is installed and we're not in CI. To be clear, the tests will be never skipped in GitHub Actions and all other environments that set `os.environ["CI"] = true`.

I also ran

```sh
pre-commit autoupdate
pre-commit run -all-files
```

since the `black==22.1.0` hook was giving me errors and aborting commits.


janosh:
The main goal is to minimize setup and keep it as simple as possible for people not changing the Q-Chem module to contribute and test their changes.

Before this PR and without installing `openbabel`, `pytest custodian` would raise

```py
E   RuntimeError: ERROR: Openbabel must be installed in order to use Q-Chem Custodian!
__ ERROR collecting test_files/qchem/new_test_files/FF_working/test_cust.py __
ImportError while importing test module
```

janosh:
The only change to review here is 03d2004. Everything else `pre-commit` is auto-formatting.

PR #228, opened 2022-08-09
Added CONTCAR to POSCAR copy for some electronic convergence errors

MichaelWolloch:
I noticed that for some troublesome relaxations, I would have electronic issues along the way,
mainly `edddav` and `eddrmm` errors. They did not occur immediately, but after potentially many ionic steps.

## Summary

* With this change to the `VaspErrorHandler` class I propose to copy CONTCAR to POSCAR (as it is done already for problems with ionic algorithms) also for electronic algorithm issues.
* However, it is important to check whether there is a valid CONTCAR to copy, which is not the case if the error occurs in the first iteration. Thus I check if the CONTCAR can be used to make a `pymatgen.io.vasp.inputs.Poscar` object, and only copy if that is the case.
* eddrmm, edddav, and zheev errors receive all the same treatment.

## TODO

* I have not yet written any unit tests for this, since I am not sure if this change is wanted. If the developers think this makes sense, I am happy to write some tests.


shyuep:
Thanks. Can you add a few lines to the [test_eddrmm](https://github.com/materialsproject/custodian/blob/048adec8c965c1a01d69ac8294f039638941df74/custodian/vasp/tests/test_handlers.py#L285) and [test_edddav]( https://github.com/materialsproject/custodian/blob/048adec8c965c1a01d69ac8294f039638941df74/custodian/vasp/tests/test_handlers.py#L302) to check that the CONTCAR copy procedure is correctly done? Thanks.

MichaelWolloch:
Hi @shyuep , I added a new CONTCAR.eddav_eddrm file that is slightly different to the POSCAR in the test folder. And added some checks to the unit test. I am not very experienced with unit tests however, and thus not sure if what I did is sufficient.

Also, should I save the original POSCAR and copy it back at the end of the test? Or is that done automatically?

shyuep:
Thanks. It should be ok.

PR #231, opened 2022-09-15
Bugfix for job termination when multiple VaspJobs are running on a single node

MichaelWolloch:
## Summary

To date, custodian terminates all VASP instances on a node when it wants to make a correction by using `os.system(f"killall {k}")`, (where k are all strings in vasp_cmd that contain "vasp". This is no problem if only one `VaspJob` is running on the machine, but if there are more, e.g. because one is using `rlaunch multi` in FireWorks, all get killed.

This fix updates the termination of processes to be specific to a process id group, and sends a `SIGTERM` signal for the termination.

Necessary modifications:

* The `subprocess.Popen()` calls in `VapJob` and `VaspNEBJob` do now set `start_new_session=True` so setsid() system calls will be made in the child process prior to the execution of the subprocess.
* The `subprocess.Popen()` is not directly returned, but saved as a class property, so the `terminate` method has access to the pid.
* The terminate method has been completely changed to replace `killall` with `os.killpg(os.getpgid(self.sbprcss.pid), signal.SIGTERM)`, which will only target the classes own sub-process group.

## Possible changes

* I am not sure if the changes to `VaspNEBJob` are actually necessary, since this class has no terminate method. But I think it cannot hurt to have the object saved in the class.

### DIFF
--- a/custodian/vasp/jobs.py
+++ b/custodian/vasp/jobs.py
@@ -257,7 +257,7 @@ class VaspJob(Job):
             self.sbprcss = subprocess.Popen(
                 cmd, stdout=f_std, stderr=f_err, start_new_session=True
             )  # pylint: disable=R1732
-            return self.sbrpcss
+            return self.sbprcss
 
     def postprocess(self):
         """
@@ -857,7 +857,7 @@ class VaspNEBJob(VaspJob):
             self.sbprcss = subprocess.Popen(
                 cmd, stdout=f_std, stderr=f_err, start_new_session=True
             )  # pylint: disable=R1732
-            return self.sbrpcss
+            return self.sbprcss
 
     def postprocess(self):
         """
--- a/custodian/vasp/jobs.py
+++ b/custodian/vasp/jobs.py
@@ -151,6 +151,7 @@ class VaspJob(Job):
         self.gamma_vasp_cmd = gamma_vasp_cmd
         self.copy_magmom = copy_magmom
         self.auto_continue = auto_continue
+        self.sbprcss = None
 
         if SENTRY_DSN:
             # if using Sentry logging, add specific VASP executable to scope
@@ -765,6 +766,7 @@ class VaspNEBJob(VaspJob):
         self.settings_override = settings_override
         self.neb_dirs = []  # 00, 01, etc.
         self.neb_sub = []  # 01, 02, etc.
+        self.sbprcss = None
 
         for path in os.listdir("."):
             if os.path.isdir(path) and path.isdigit():
--- a/custodian/vasp/jobs.py
+++ b/custodian/vasp/jobs.py
@@ -253,10 +253,9 @@ class VaspJob(Job):
         logger.info(f"Running {' '.join(cmd)}")
         with open(self.output_file, "w") as f_std, open(self.stderr_file, "w", buffering=1) as f_err:
             # use line buffering for stderr
-            self.sbprcss = subprocess.Popen(cmd,
-                                            stdout=f_std,
-                                            stderr=f_err,
-                                            start_new_session=True)  # pylint: disable=R1732
+            self.sbprcss = subprocess.Popen(
+                cmd, stdout=f_std, stderr=f_err, start_new_session=True
+            )  # pylint: disable=R1732
             return self.sbrpcss
 
     def postprocess(self):
@@ -853,10 +852,9 @@ class VaspNEBJob(VaspJob):
         logger.info(f"Running {' '.join(cmd)}")
         with open(self.output_file, "w") as f_std, open(self.stderr_file, "w", buffering=1) as f_err:
             # Use line buffering for stderr
-            self.sbprcss = subprocess.Popen(cmd,
-                                            stdout=f_std,
-                                            stderr=f_err,
-                                            start_new_session=True)  # pylint: disable=R1732
+            self.sbprcss = subprocess.Popen(
+                cmd, stdout=f_std, stderr=f_err, start_new_session=True
+            )  # pylint: disable=R1732
             return self.sbrpcss
 
     def postprocess(self):
--- a/custodian/vasp/jobs.py
+++ b/custodian/vasp/jobs.py
@@ -254,9 +254,9 @@ class VaspJob(Job):
         with open(self.output_file, "w") as f_std, open(self.stderr_file, "w", buffering=1) as f_err:
             # use line buffering for stderr
             self.sbprcss = subprocess.Popen(cmd,
-                                      stdout=f_std,
-                                      stderr=f_err,
-                                      start_new_session=True)  # pylint: disable=R1732
+                                            stdout=f_std,
+                                            stderr=f_err,
+                                            start_new_session=True)  # pylint: disable=R1732
             return self.sbrpcss
 
     def postprocess(self):
@@ -854,9 +854,9 @@ class VaspNEBJob(VaspJob):
         with open(self.output_file, "w") as f_std, open(self.stderr_file, "w", buffering=1) as f_err:
             # Use line buffering for stderr
             self.sbprcss = subprocess.Popen(cmd,
-                                      stdout=f_std,
-                                      stderr=f_err,
-                                      start_new_session=True)  # pylint: disable=R1732
+                                            stdout=f_std,
+                                            stderr=f_err,
+                                            start_new_session=True)  # pylint: disable=R1732
             return self.sbrpcss
 
     def postprocess(self):
--- a/custodian/vasp/jobs.py
+++ b/custodian/vasp/jobs.py
@@ -5,6 +5,7 @@ This module implements basic kinds of jobs for VASP runs.
 import logging
 import math
 import os
+import signal
 import shutil
 import subprocess
 from shutil import which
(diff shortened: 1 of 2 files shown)


MichaelWolloch:
The failed linting does not make too much sense to me. .sbprcss should be always defined... Help is appreciated there.

shyuep:
Thanks.

matthewkuner:
@MichaelWolloch is it possible that this might prevent some jobs from getting terminated that should be? I am running into an issue on one of NSF's supercomputers where custodian isn't killing + restarting a job when a handler catches an error

MichaelWolloch:
Hi @matthewkuner !
I really cannot tell without much more information. As far as I know that should not happen, but then I did test this only on a limited  number of systems.

However, this has been merged for a while now, and if this would be a common problem, more people would have complained I think.

I would suggest to maybe test this out in an interactive session on your cluster. Run a script like the one below in a folder with VASP inputs for a quick calculation and check if VASP is stopped after 10 seconds. Also make sure that there are some correct output lines in `vasp.out` and the `vasp.err` file has lines like this:
`forrtl: error (69): process interrupted (SIGINT)`.

~~~
#!/usr/bin/env python3
# -*- coding: utf-8 -*-
"""
Created on Thu Apr 20 10:34:00 2023

@author: wolloch
"""
import os
import signal
import subprocess
import time

cmd=['mpirun', '-np', '4', 'vasp.6.4.1_std']

def run_vasp():
    with open('vasp.out', "w") as f_std, open('vasp.err', "w", buffering=1) as f_err:
            # use line buffering for stderr
            sbprcss = subprocess.Popen(
                cmd, stdout=f_std, stderr=f_err, start_new_session=True
            )  # pylint: disable=R1732
            return sbprcss
        
def terminate_vasp(sbprcss):
    try:
        os.killpg(os.getpgid(sbprcss.pid), signal.SIGTERM)
    except Exception as e: # work on python 3.x
        print(f'Exception:\n {e}')
        
if __name__ == "__main__":
    
    vasp_process = run_vasp()
    
    time.sleep(10)
    
    terminate_vasp(vasp_process)
~~~

matthewkuner:
Thank you for the help (and quick response time)! I ran the code snippet you provided, and got `forrtl: error (78): process killed (SIGTERM)` (same message, though different error number), and the vasp.out looked normal. Hmm. 
Below is the text that was outputted to vasp.err:

> Image              PC                Routine            Line        Source
vasp_std           0000000001E7F40B  Unknown               Unknown  Unknown
libpthread-2.28.s  000014A9C89BDDD0  Unknown               Unknown  Unknown
libmpi.so.12.0.0   000014A9C920B93F  Unknown               Unknown  Unknown
libmpi.so.12.0.0   000014A9C91FFF87  Unknown               Unknown  Unknown
libmpi.so.12.0.0   000014A9C917B6C3  Unknown               Unknown  Unknown
libmpi.so.12.0.0   000014A9C9152C51  Unknown               Unknown  Unknown
libmpi.so.12.0.0   000014A9C9136ACA  Unknown               Unknown  Unknown
libmpi.so.12.0.0   000014A9C9219DE5  Unknown               Unknown  Unknown
libmpi.so.12.0.0   000014A9C9115D6D  MPI_Barrier           Unknown  Unknown
libmpifort.so.12.  000014A9CA4B77BC  pmpi_barrier          Unknown  Unknown
vasp_std           000000000047A603  Unknown               Unknown  Unknown
vasp_std           0000000001BA5616  Unknown               Unknown  Unknown
vasp_std           0000000001C17A1F  Unknown               Unknown  Unknown
vasp_std           000000000040C222  Unknown               Unknown  Unknown
libc-2.28.so       000014A9C828A6A3  __libc_start_main     Unknown  Unknown
vasp_std           000000000040C12E  Unknown               Unknown  Unknown


Perhaps something is out of the ordinary here? Unsure.

matthewkuner:
Oh, I just discovered another weird bug. From one of my runs that wasn't getting restarted properly, the vasp.out file after the first error just looks like 

> RMM:   7    -0.627876212974E+03   -0.56620E-02   -0.92683E-02 54414   0.391E-01    0.256E+00
RMM:   8    -0.627874473543E+03    0.17394E-02   -0.45164E-02 54242   0.270E-01    0.227E+00
RMM:   9    -0.627875858999E+03   -0.13855E-02   -0.26367E-02 53814   0.176E-01    0.202E+00
RMM:  10    -0.627879585529E+03   -0.37265E-02   -0.19599E-02 49789   0.196E-01    0.130E+00
RMM:  11    -0.627882081610E+03   -0.24961E-02   -0.27710E-02 52737   0.305E-01    0.140E+00
RMM:  12    -0.627877258320E+03    0.48233E-02   -0.16716E-02 53548   0.240E-01    0.142E+00
RMM:  13    -0.627875092709E+03    0.21656E-02   -0.55086E-03 48747   0.126E-01    0.988E-01
RMM:  14    -0.627875085091E+03    0.76174E-05   -0.32416E-03 49930   0.611E-02    0.681E-01
RMM:  15    -0.627875083487E+03    0.16046E-05   -0.14938E-03 52644   0.404E-02    0.495E-01
RMM:  16    -0.627875006497E+03    0.76990E-04   -0.13042E-03 50601   0.544E-02    0.119E-01
RMM:  17    -0.627875113968E+03   -0.10747E-03   -0.98323E-04 52485   0.507E-02    0.172E-01
RMM:  18    -0.627875060335E+03    0.53633E-04   -0.43912E-04 45057   0.340E-02    0.723E-02
RMM:  19    -0.627875053655E+03    0.66796E-05   -0.26751E-04 46848   0.188E-02    0.630E-02
RMM:  20    -0.627875068529E+03   -0.14874E-04   -0.16121E-04 50040   0.120E-02    0.629E-02
RMM:  21    -0.627875070943E+03   -0.24138E-05   -0.83724E-05 49717   0.841E-03    0.466E-02
RMM:  22    -0.627875075203E+03   -0.42599E-05   -0.57129E-05 46023   0.599E-03    0.146E-02


with none of the earlier text typically present...

matthewkuner:
Question:

I go into each directory error.1.tar.gz, error.2.tar.gz, etc., and I look at the OUTCAR that is saved for each of those error checkpoints. I should see a different runtime at the top for each of them, correct? And the INCAR parameters printed to the OUTCAR should be different (assuming they were changed by some error handler). Correct?

matthewkuner:
Hmm, it's good to know that the jobs seem to be getting killed normally / that the weird symbols probably aren't significant. 

I wonder if using the "update_user_incar_settings" function in Atomate2 might cause this. Hmm.

This issue is actually reproducible for me, though perhaps I'll just have to change my initial INCAR parameters to circumvent this for now...

matthewkuner:
Update: I modified the custodian.vasp.jobs.py file to contain the following:

>     def terminate(self):
        """
        Kill all vasp processes associated with the current job.
        """
        logger.info(f"Custodian terminating all VASP processes within process group {self.sbprcss.pid}")
        try:
            print(os.getpgid(self.sbprcss.pid))
            os.killpg(os.getpgid(self.sbprcss.pid), signal.SIGTERM)
            # write a file signifying this was attempted
            with open('termination_was_attempted.txt', 'w') as fp:
                pass
        except Exception as e:
            with open('termination_failed.txt', 'w') as f:
                f.write(str(e))
            print('something went wrong in trying to kill the jobs')
            pass

and when I ran a real firework, the termination failed and the "termination_failed.txt" file contained the error message "[Errno 3] No such process"

matthewkuner:
On a different supercomputer, I just updated custodian to a version that has this bugfix, and got this error for the first time on that supercomputer. So I think there may be something wrong here? Or, perhaps it is related to how this interfaces with Atomate2? Unsure

matthewkuner:
I just modified my `custodian.vasp.jobs.py` (which is from the latest version of custodian) file to contain the old method for terminating jobs, and it seems to be functioning correctly with Atomate2 (also the latest version). 

In case it matters, these issues I've been having were on Savio and on NSF ACCESS's Bridges2 (not NERSC--I haven't ran jobs there in a few weeks due to maintenance, etc.)

MichaelWolloch:
Hi again, and thanks for all the tests. 

Seems strange. Especially since the test script did work correctly. Are you running more than one job per node? If not, than the old version should be fine of course, but it is still a problem that the new version is failing. 

What does your logfile show? Does the following line show a process, or is already this failing?
`logger.info(f"Custodian terminating all VASP processes within process group {self.sbprcss.pid}")`

matthewkuner:
I'm running one job across 8 nodes. Regarding the logfile--to be honest, I don't actually know where to check that. I've looked at the atomate2/logs directory, but nothing of use is written there in my experience.

matthewkuner:
If my memory serves me well, I believe the `self.sbprcss.pid` variable was set--it was a single number, I believe.

MichaelWolloch:
I think this could have something to do with running on more than one node. In principle this should still work, but I am not sure if I tested it.
I think I took care to kill the whole process group, not a single process. But it is conceivable that the process running custodian is running on a different node than the process it wants to kill. Thus it has the pid number, but when you try to kill it, it does not exist.

Can you try and run the test script with your modifications (maybe in a loop to make it unlikely to get lucky) on a VASP job that runs on 8 nodes?

I will also investigate a bit more, but right now I am pretty swamped.

I usually run a lot of smaller jobs on big nodes. That is the reason why I changed this in the first place, because one custodian restart would kill all of my jobs on the node.

matthewkuner:
@MichaelWolloch I haven’t had the chance to do any more tests. Nonetheless, I think a reasonable solution would be to change the terminate() function to be:
```
def terminate_vasp(sbprcss):
    try:
        # try to kill based on sbprcss.pid
        try:
            os.killpg(os.getpgid(sbprcss.pid), signal.SIGTERM)
        # try to kill using "killall"
        except:
            cmds = self.vasp_cmd
            if self.gamma_vasp_cmd:
                cmds += self.gamma_vasp_cmd
            for k in cmds:
                if "vasp" in k:
                    try:
                        os.system(f"killall {k}")
    except Exception as e: # work on python 3.x
        print(f'Exception:\n {e}')
```

Let me know what you think

PR #233, opened 2022-11-01
SCF Handler Patch

nwinner:
Previous SCF handler assumed that input file would be present and attribute "is_ot" could be defined. This is not actually true, and the handler is initialized before the input file is actually written. Instead, check for is_ot during the handler check method. Extraneous info that was collected by assuming the input file has been removed.


nwinner:
@shyuep I added a test that the initialization of all the handlers goes over fine. I also threw in another change where I dropped the basis set adjustment in the handlers. Full control of fixing basis sets during SCF handling is way more complex than what was in the previous version. What the handler was doing was really just input set generation by setting preferred basis set. So that should go into pmg + atomate(2) input set generation settings.

PR #243, opened 2023-05-06
VASP Handler: Do not set `ADDGRID` to True

### LINKED ISSUE #226: Use of ADDGRID in VASP Custodian should be investigated
As noted in the [VASP manual](https://www.vasp.at/wiki/index.php/ADDGRID), when using ADDGRID, the VASP team "recommend[s] performing careful tests, on whether ADDGRID works as desired; please do not use this tag as default in all your calculations!" Empirically, I have found that `AECCAR` files oftentimes are not valid when `ADDGRID` is set to True, which can influence partial atomic charge calculations (and related methods like bond order analyses) downstream. This begs the question of whether `ADDGRID` should be used at all in Custodian. This should be investigated and reconsidered. Personally, I am currently leaning towards removing it altogether from Custodian. It's not used in many places as-is.

arosen93:
This closes #226. In short, the VASP manual strongly advises caution against using the `ADDGRID` command, and in my experience, I have seen that it produces invalid `AECCAR` files. It should not be set in Custodian as a rule.

### DIFF
--- a/custodian/vasp/handlers.py
+++ b/custodian/vasp/handlers.py
@@ -898,11 +898,8 @@ class DriftErrorHandler(ErrorHandler):
         # Move CONTCAR to POSCAR
         actions.append({"file": "CONTCAR", "action": {"_file_copy": {"dest": "POSCAR"}}})
 
-        # First try adding ADDGRID
-        if not incar.get("ADDGRID", False):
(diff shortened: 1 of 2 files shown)


arosen93:
@shyuep this is ready to merge.

shyuep:
Thanks!

PR #244, opened 2023-05-06
VASP Handler: AMIN handler should only be applied if SCF is not converged

### LINKED ISSUE #229: AMIN error handler unnecessarily applied
Sometimes VASP prints the following warning for large cells:
```
|     One of the lattice vectors is very long (>50 A), but AMIN is rather     |
|     large. This can spoil convergence since charge sloshing might occur     |
|     along the long lattice vector. If problems with convergence are         |
|     observed, try to decrease AMIN to a smaller value (e.g. 0.01).          |
|     Note: This warning only applies if the self-consistency cycle is        |
|     used.                                                                   |
```

This is printed whether or not the system fails to converge and is only a warning that the system MIGHT not converge.

However, the "amin" error handler in custodian is applied whether or not there was a problem with convergence. This means, calculations that still converged fine are thrown out at the end, leading to a waste of computer time.

Can the amin handler be updated so that it is only applied if there are also convergence issues?

Alternatively, since this error message is printed immediately before the SCF cycle, can the error message be caught immediately before any time is wasted on the calculation.

arosen93:
Closes #229. The AMIN handler should only be applied if SCF convergence is not achieved.


arosen93:
@shyuep --- this one should be ready to go.

PR #245, opened 2023-05-06
VASP Handler: Only apply the algo_tet handler if SCF convergence failed

### LINKED ISSUE #221: Suggestion: disable algo_tet VASP handler by default
#180 and #202 introduced an error handler to catch calculations with `ALGO = All` and `ISMEAR = -5` based on a warning in the OUTCAR file.

Currently, this error handler means it is a pain to run high quality hybrid density of states (you need to create a pre-converged WAVECAR) and then run `ALGO=DAMPED` + tetrahedron method.

I'm curious if this error handler was added due to experience or just based on the warning itself. In my experience running hybrid of states calculations (on 300+ systems) I've never seen this warning lead to a crash.

I would suggest disabling this error message by default. If we want to keep it enabled, the fix should only be applied if the calculation actually failed and not if the calculation completes successfully as is currently the case.

@arosen93 what are your thoughts?

arosen93:
Closes #221. The `algo_tet` handler in VASP currently makes hybrid DOS calculations impractical. Since the handler is based on a warning, it should only activate if SCF convergence is not achieved.


arosen93:
@utf: Let me know if this isn't to your liking. I'll add tests soon.

arosen93:
@shyuep --- this is ready to go.

shyuep:
There seems to be merge conflicts introduced from merging the other PRs. While I can theoretically fix this on my end, I think it might be better for you to do it to make sure it is correct done. 

Thanks for all the contributions!

arosen93:
No problem. I was aware this would happen! I'll patch them this afternoon.

arosen93:
@shyuep --- this one is now ready again. Thanks for merging!

PR #248, opened 2023-05-06
VASP Handler: Remove deprecated handlers and kwargs

arosen93:
This PR removes handlers that have aged past their deprecation time. There is no attached issue.

### DIFF
--- a/custodian/vasp/handlers.py
+++ b/custodian/vasp/handlers.py
@@ -108,7 +108,6 @@ class VaspErrorHandler(ErrorHandler):
     def __init__(
         self,
         output_filename="vasp.out",
-        natoms_large_cell=None,
         errors_subset_to_catch=None,
         vtst_fixes=False,
     ):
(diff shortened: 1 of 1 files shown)


arosen93:
@shyuep this is ready to merge.

mkhorton:
Removing a keyword argument (`natoms_large_cell`) causes issues with already-serialized Custodian objects, e.g. attached to workflows, especially since the `VaspErrorHandler` doesn't accept arbitrary keyword arguments.

mkhorton:
i.e., this PR is a breaking change. I'm not sure the best resolution, but I imagine it will cause issues for any existing users of custodian with large databases that encode the specific error handlers they use, leading to de-serialization errors or an inability to upgrade custodian (a minimal fix could just be to add `**kwargs` to `VaspErrorHandler.__init__` as a no-op).

arosen93:
Good point. Alright, I can just as easily put it back! Easy enough. I'll open a PR.

mkhorton:
It’s not the most elegant solution, but maybe just the `**kwargs` addition is sufficient, rather than adding it all back: this way we still remove the old, deprecated code, but do not encounter de-serialisation issues.

PR #249, opened 2023-05-07
VASP Handler: General clean up and mitigating unnecessary INCAR swaps

arosen93:
In this PR, I have done a lot of cleaning up of the VASP `handlers.py` file, which was getting unwieldy due to lots of inconsistencies between contributors. Functionally, there are minimal differences --- the only noteworthy change is that I now check the default values of the INCAR flag before potentially updating it. For instance, in an error that's due to SYMPREC being smaller 1e-4, we don't need to set SYMPREC to 1e-4 if SYMPREC is already 1e-3. In addition to those kinds of checks, I also collapsed nested if statements that did not need to be nested.

To emphasize that no substantial changes to the code execution were made, I have left the unit tests in tact. The only test I removed was one that did not seem to belong and was a duplicate.


arosen93:
@shyuep --- this is ready to merge.

PR #250, opened 2023-05-07
Fix recent merge conflict due to incorrect merging of test suite

arosen93:
Fixes the recent issue caused by a faulty merge conflict in master. Waiting on the tests this time, but locally we are good...

For context, there are now accidentally two algo_tet error handlers in the test suite, but only one was supposed to be kept (the one under the UnconvergedErrorHandler). The other one was supposed to be deleted when the new one was added.


arosen93:
@shyuep --- okay _now_ we are good to go. Sorry 'bout that!

PR #254, opened 2023-05-18
Change VASP terminate function to use the old `killall` method as a backup

matthewkuner:
## Summary

@shyuep This adds in a backup method for terminating VASP jobs. Specifically, if the recent method added in #231 does not work, try to use the old method (based on `killall`) as a backup.

Reason for this change: the newer method added in #231 did not work in many of my test cases on the NSF ACCESS' Bridges2 supercomputer or Berkeley's Savio supercomputer. The old method did, though is not preferable in cases where multiple jobs are running on a single node. Hence, using the old method as a backup seems sensible to me.

### DIFF
--- a/custodian/vasp/jobs.py
+++ b/custodian/vasp/jobs.py
@@ -673,7 +673,16 @@ class VaspJob(Job):
         """
         logger.info(f"Custodian terminating all VASP processes within process group {self.sbprcss.pid}")
         try:
-            os.killpg(os.getpgid(self.sbprcss.pid), signal.SIGTERM)
+            try: # first try to terminate all VASP processes based on sbprcss.pid
+                os.killpg(os.getpgid(self.sbprcss.pid), signal.SIGTERM)
+            except Exception: # As a backup method, try to kill using "killall"
+                cmds = self.vasp_cmd
+                if self.gamma_vasp_cmd:
+                    cmds += self.gamma_vasp_cmd
+                for k in cmds:
+                    if "vasp" in k:
+                        try:
+                            os.system(f"killall {k}")
         except Exception:
             pass
 
--- a/custodian/vasp/jobs.py
+++ b/custodian/vasp/jobs.py
@@ -673,15 +673,15 @@ class VaspJob(Job):
         """
         logger.info(f"Custodian terminating all VASP processes within process group {self.sbprcss.pid}")
         try:
-            try: # first try to terminate all VASP processes based on sbprcss.pid
+            try:  # first try to terminate all VASP processes based on sbprcss.pid
                 os.killpg(os.getpgid(self.sbprcss.pid), signal.SIGTERM)
-            except Exception: # As a backup method, try to kill using "killall"
+            except Exception:  # As a backup method, try to kill using "killall"
                 cmds = self.vasp_cmd
                 if self.gamma_vasp_cmd:
                     cmds += self.gamma_vasp_cmd
                 for k in cmds:
                     if "vasp" in k:
-                    	try:
+                        try:
                             os.system(f"killall {k}")
                         except Exception:
                             pass
--- a/custodian/vasp/jobs.py
+++ b/custodian/vasp/jobs.py
@@ -681,8 +681,10 @@ class VaspJob(Job):
                     cmds += self.gamma_vasp_cmd
                 for k in cmds:
                     if "vasp" in k:
-                        try:
+                    	try:
                             os.system(f"killall {k}")
+                        except Exception:
+                            pass
         except Exception:
             pass
 


arosen93:
Personally, I think if the changed introduced in #231 is causing issues, we should consider either correcting it or reverting that change. The need to run Custodian when jobs are using multiple nodes is much greater than the edge case where the user is trying to run multiple jobs on a single node. It seems to me that basically always having a `try`/`except` fail except in edge cases is a bit extreme. The old method worked fine, and it's clear the new one does not based on your thorough testing.

@shyuep, thoughts on this?

custodian/vasp/jobs.py
```diff
@@ -673,7 +673,18 @@ def terminate(self):
         """
         logger.info(f"Custodian terminating all VASP processes within process group {self.sbprcss.pid}")
         try:
-            os.killpg(os.getpgid(self.sbprcss.pid), signal.SIGTERM)
+            try:  # first try to terminate all VASP processes based on sbprcss.pid
```
janosh, line 676: I don't think we need 2 levels of `try` here. There's an unused except case below.

janosh:
I'm usually quick to dismiss security concerns in research code but wanted to at least mention that `os.system()`, especially since the command is constructed with string formatting here, is susceptible to shell injection attacks. Would be better practice to use `subprocess.run()`, which takes a a list of args, not a single string.

custodian/vasp/jobs.py
matthewkuner: I agree—I was just matching the unused `except` from the original code (which was present both before and after #231 )

MichaelWolloch:
Hi, In principle the best thing would be to find a solution that is working for all use cases. I agree that it is more important to have this work on multiple node jobs, but I also think that the other (my) use case of multiple jobs on a single node will increase when  nodes get larger and larger. For example on the cluster I am mostly using each node has 2 EPYC 64 cores installed, and for some reason VASP only runs well for up to 64 cores per node, but scales well over multiple  nodes.

Option 2 of @arosen93 is possible of course, but better would be to patch #231 so that it works in both cases. I probably have some time next week to spend on this, but now that more people are involved, I am quite sure that there will be more ideas.

However, just reverting #231 is also fine as a stopgap of course.

MichaelWolloch:
Hello again,
I have to admit that I know way to little about how processes are distributed across nodes and who knows what is where. But somehow, `killall` when called via `os.system()` works, while  `os.killpg(os.getpgid(self.sbprcss.pid), signal.SIGTERM)` does not.

Could it be that the kill command would work? So, also including @janosh's security concerns, something like this:
`kill = subprocess.Popen(['kill', f'{-os.getpgid(sbprcss.pid)}'])`

if you pass a negative PID to kill, it kills all processes in that process group according to the [man page](https://man7.org/linux/man-pages/man2/kill.2.html). My, probably naive, idea is that the shell "knows" more about the processes than python and the clusters process management can work its magic across multiple nodes then.

I have tested this version on my desktop right now, and it does not kill other vasp processes running at the same time, but correctly kills the ones associated with the PID passed. Of course that was also true previously.

Maybe @matthewkuner could test this approach on a job using mutliple nodes? I will also try to test it on my cluster, but I hope that Matthew has some test jobs (maybe even using custodian) ready to go.

arosen93:
Interesting finding, @MichaelWolloch! I do agree with both of you here that a real solution would be ideal if we can find one!

I don't have any good minimal examples that rely on Custodian error handling but hopefully @matthewkuner you might be able to test the idea out when you have a chance? If not, let me know and I can try to whip something up later.

MichaelWolloch:
I just set up a quite minimal script and just run it using a "bad input" (minimum number of bands, tetrahedral smearing for metal, small POTIM, large SIGMA, not enough NSW, etc...) in a job on two nodes. I.e. I am just putting the INCAR, POSCAR, KPOINTS, and POTCAR in the directory and run VASP through custodian with this script which I execute in the job file.
~~~
from custodian.custodian import Custodian
from custodian.vasp.jobs import VaspJob
from custodian.vasp.handlers import (VaspErrorHandler,
                                     UnconvergedErrorHandler,
                                     PotimErrorHandler,
                                     LargeSigmaHandler)                                     
                                     


cmd=['mpirun','-np','96','vasp.6.4.1_std']
handlers = [VaspErrorHandler(),
            UnconvergedErrorHandler(),
            PotimErrorHandler(),
            LargeSigmaHandler()]

job = VaspJob.double_relaxation_run(vasp_cmd=cmd,half_kpts_first_relax=True)
c = Custodian(handlers, job, max_errors=10)
c.run()
~~~
The more people to try this on different clusters, the better. The cluster I am on now is using slurm 22.05.2, and runs on AlmaLinux:
~~~
NAME="AlmaLinux"
VERSION="8.5 (Arctic Sphynx)"
ID="almalinux"
ID_LIKE="rhel centos fedora"
VERSION_ID="8.5"
PLATFORM_ID="platform:el8"
PRETTY_NAME="AlmaLinux 8.5 (Arctic Sphynx)"
ANSI_COLOR="0;34"
CPE_NAME="cpe:/o:almalinux:almalinux:8::baseos"
HOME_URL="https://almalinux.org/"
DOCUMENTATION_URL="https://wiki.almalinux.org/"
BUG_REPORT_URL="https://bugs.almalinux.org/"

ALMALINUX_MANTISBT_PROJECT="AlmaLinux-8"
ALMALINUX_MANTISBT_PROJECT_VERSION="8.5"
~~~

arosen93:
Thanks!! I will try it out today!

MichaelWolloch:
OK, I played around with this a bit because I am not obsessed and cannot concentrate on the paper I am supposed to write. I have not solved it, but I am a lot closer to understand what is going on:

1. In my new implementation I tried to kill all vasp processes that belong to a certain vasp instance. For that I tried to use process group IDs.
2. In fact, I cannot set the process group IDs for two reasons:
    - I do not have the permissions to use [setpgid](https://man7.org/linux/man-pages/man2/setpgid.2.html) (via the `process_group` input argument of `subprocess.Popen()`
    - I am not actually creating those subprocess in python, mpirun is creating them!
3. So what I actually did is killing all process with the process group IDs of the `mpirun` process. Which is just one process, the `mpirun` process. It is the mpirun process for the correct vasp instance however, which is good.
4. This should not actually be that bad, since if you terminate mpirun, it should send SIGTERM to all the mpi processes, much like hitting `Ctrl-C` on the keyboard. And this does seem to work on a single node or workstation without issues.
5. I do not yet understand why SIGTERM is not correctly propagated to processes running on the other nodes.
6. It may be that there is some involvement from the scheduler on the cluster, or rather some environment variables set by the scheduler that are used by mpirun to communicate with the other nodes.
7. I assume that when you terminate the vasp processes directly (via killall), that the processes on the other nodes are also terminated due to

arosen93:
What happens if the user is instead using `srun` instead of `mpirun`? Will this be transferable? e.g. https://docs.nersc.gov/applications/vasp/

MichaelWolloch:
OK, I think I figured this out. But I am waiting for my test job to start so I am not fully sure.

As I previously wrote, the issue seems to be that the mpirun (or srun, etc) command gets killed instead of the vasp processes, which are children of this command.

One has access to this [children](https://psutil.readthedocs.io/en/latest/index.html?highlight=children#psutil.Process.children) if one uses the `psutil` package. Then one can search for the processes containing "vasp" in those children and kill those. (Note that if someone names his vasp executable in a way that does not include vasp, they are out of luck. But that was the case previously as well with the `killall` implementation.)

E.g. `mpirun -np 4 vasp.6.4.1_std` will spawn `mpiexec.hydra`, which will spawn `hydra_pmi_proxy`, which then will spawn finally 4 vasp processes. If one uses the recursive option of the children() command of a psutil.Process, one can list them all with their respective PIDs and other info:
~~~
[psutil.Process(pid=11314, name='mpiexec.hydra', status='sleeping', started='17:16:32'),
 psutil.Process(pid=11316, name='hydra_pmi_proxy', status='sleeping', started='17:16:32'),
 psutil.Process(pid=11321, name='vasp.6.4.1_std', status='running', started='17:16:32'),
 psutil.Process(pid=11323, name='vasp.6.4.1_std', status='running', started='17:16:32'),
 psutil.Process(pid=11325, name='vasp.6.4.1_std', status='running', started='17:16:32'),
 psutil.Process(pid=11327, name='vasp.6.4.1_std', status='running', started='17:16:32')]
~~~
My proposed terminate function then simply tries to kill these processes one by one and raise meaningful exceptions if that does not work:
~~~
    def terminate(self):
        """
        Kill all vasp processes associated with the current job.
        This is done by looking for child processes of the subprocess that
        runs vasp (e.g. mpirun)
        """
        logger.info(f"Custodian terminating all VASP processes that are "
                    f"children of {self.sbprcss.pid}")
        child_processes = self.sbprcss.children(recursive=True)
        for child in child_processes:
            if 'vasp' in child.name().lower():
                try:
                    os.kill(child.pid,signal.SIGTERM)
                except ProcessLookupError:
                    logger.warning(f"Failed to kill process {child.name()} "
                          "with pid {child.pid}: Process not found.")
                except PermissionError:
                    logger.warning(f"Failed to kill process {child.name()} "
                          "with pid {child.pid}: Permission denied.")
                except Exception as e:
                    logger.warning(f"Failed to kill process {child.name()} "
                          f"with pid {child.pid}: {str(e)}")
~~~
On my workstation that works well, and kills only the vasp instance it is supposed to. Since it kills vasp processes directly instead of the mpi executable, it should also work accross many nodes, like `killall`.

@janosh , I think `os.kill` should be save, right? Only `os.system` was fishy?

I can preprare a PR for this...

janosh:
Nice!

> I can prepare a PR for this...

@MichaelWolloch I think that would be great! This seems safer.

Re `ProcessLookupError`:

```py
            if 'vasp' in child.name().lower():
                try:
                    os.kill(child.pid,signal.SIGTERM)
                except ProcessLookupError:
                    logger.warning(f"Failed to kill process {child.name()} "
                          "with pid {child.pid}: Process not found.")
```

would it make sense to check that the process wasn't already killed? Maybe add [`and child.is_running()`](https://psutil.readthedocs.io/en/latest/index.html?highlight=is_running#psutil.Process.is_running) to the `if 'vasp' in child.name().lower()` check. Might give fewer spurious warnings in the job logs. Chunky job logs full of warnings are one of my pet peeves. 😆

janosh:
Just to add background, I was thinking of race conditions that often arise in concurrent multi-process jobs, i.e. where another process or thread might have already killed the same set of children.

MichaelWolloch:
Hello again everybody.

Unfortunately, the idea with the child processes is not working when job schedulers are involved, because they interfere with the parent-children chain and now the vasp processes are no longer in the child-list of the mpirun.

However, I think I got a new and good alternative. To remind everybody, it is necessary to kill at least one vasp mpi process, which will send a kill signal to all the others (also on other nodes). At the same time, we should not kill all vasp processes, because many jobs might run on one node. There is even a [mode](https://materialsproject.github.io/fireworks/multi_job.html) for that in FireWorks.

The new idea looks for processes that access the same OUTPUT files and kills them. The directory is easily found, because custodian runs in the execution directory. So one can loop over all processes, and select those that are called "vasp" AND access files in the current directory.
The only slight issue, is that the looping over opened files by a process ([psutil.Process.open_files()](https://psutil.readthedocs.io/en/latest/index.html?highlight=open_file#psutil.Process.open_files)) will list inaccessible files and can very easily break. One therefor needs to be careful with try/except blocks.

While it should not happen that a vasp process and its files cannot be accessed, I nevertheless included a safety, which uses a `killall` command. However, I ran two short vasp runs in parallel on my workstation for 1 and 2 seconds, respectively, and killed them correctly in a loop for 1000 times, so this should never actually happen.

A design choice is to use the `CHGCAR` file as the one that is checked. This might seem like a strange choice, compared to e.g. `OUTCAR`, but in fact, it is one of those files that gets accessed by every single vasp process, so there are possible many more attempts of killing the process.

I double checked that this also works if not CHGCAR is actually written (LCHARG = .FALSE.). The file is still created and touched by all processes. So there seems to be no issues there. However, I am open to using another file as well.

I made a PR at #259

janosh:
@MichaelWolloch Sounds great! Thanks for being so thorough! 👍

janosh:
Thank you to @matthewkuner as well! Really excited to start using your new ML pre-relax workflows in `atomate2`!

PR #256, opened 2023-05-25
VASP: Add handler for invalid WAVECAR when going from `vasp_gam` to `vasp_std`

### LINKED ISSUE #255: Add a VASP handler for when a `vasp_std` calc reads in a `vasp_gam` WAVECAR (or vice versa)
When `vasp_gam` writes a WAVECAR, it is incompatible with `vasp_std` (and vice versa). An error will occur, and the user needs to either set `ISTART = 0` or delete the `WAVECAR`/`CHGCAR` and rerun.

I didn't see a handler for this, so we should add it. I'll need to figure out the error message, but it should be easy. I'd lean towards `ISTART = 0` rather than deleting the files.

arosen93:
Closes #255.


arosen93:
@shyuep --- this one is ready to go.

Issue #258, opened 2023-05-26
Custodian doesn't exit gracefully when running VASP over multiple nodes


arosen93:
This error was introduced in a prior PR. See https://github.com/materialsproject/custodian/pull/254 for an ongoing discussion.

janosh:
@shyuep Needs closing. feel free to add me to this repo for help.

PR #259, opened 2023-06-02
Change VASP terminate function to be sensitive to execution directory

MichaelWolloch:
## Summary

This fixes #231 for multi node jobs, and should solve #254 in a better way.

* Kill vasp processes reliable on single or multi node jobs (would welcome some testing on other clusters)
* Has a `killall` backup in case that permission errors hobble the current implementation (which should not happen, since custodian also starts the vasp processes)

Unfortunately this comes with a new dependency, psutil. But is is a very small, healthy and considered a key ecosystem project.

Please see the discussion at #254 for more details.


arosen93:
Amazing! Thank you so much for your hard work on this.

arosen93:
This closes #258.

shyuep:
Looks good to me! Thanks! I will release after this.

PR #263, opened 2023-06-15
Add missing requirement to `setup.py`

arosen93:
The dependency `psutil` was included in `requirements.txt` but not in `setup.py`, causing other packages using Custodian to fail if they did not have `psutil` installed. This was a result of #259.


arosen93:
@shyuep ready to go once tests pass.

JaGeo:
@arosen93 , I ran into this issue yesterday as well. Forgot about it again... Thanks for fixing it.

shyuep:
Thanks.

PR #271, opened 2023-07-12
Fix default ISMEAR value in `LargeSigmaHandler` (should be 1, not 0)

### LINKED ISSUE #266: [BUG] Issue with LargeSigmaHandler
https://github.com/materialsproject/custodian/blob/f7dc11a15aee43e57cc8b52bf4a39ed7c5fd1e63/custodian/vasp/handlers.py#L1254C3-L1254C3

Currently, if no ISMEAR is found in the INCAR file, the LargeSigmaHandler uses a value of 0. However, the default ISMEAR from VASP is actually 1. This should be changed.

arosen93:
Closes #266.

### DIFF
--- a/custodian/vasp/handlers.py
+++ b/custodian/vasp/handlers.py
@@ -1245,7 +1245,7 @@ class LargeSigmaHandler(ErrorHandler):
             # Can't perform check if Outcar not valid
             return False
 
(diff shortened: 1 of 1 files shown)


arosen93:
@shyuep --- this one is ready to go.

Issue #268, opened 2023-07-06
Caching parsed files


gpetretto:
In most of the handlers and validators the same output files are parsed in the `check` and/or in the `correct` methods. For VASP they are mainly the vasprun.xml and OUTCAR files. In most cases this is not a huge overhead, even though it is still a bit of a waste, but for calculations with large output data (e.g. MD) the multiple parsing could have an impact.

I think it would be beneficial to modify Custodian and the handlers in such a way that the parsed data could be cached and shared among the handlers. 

If this could be interesting I can draft an implementation.  
One way of doing this could be by creating a dictionary in `Custodian._runjob()` and setting it as a `cached_data` attribute in all the handlers. Every handler could then check if a parsed version of the data is already present in the dictionary and otherwise set it. For example in a handler one could have something like:
```python
cached_data = getattr(self, "cached_data", None)
if cached_data and "Vasprun" in cached_data:
    vasprun = cached_data["Vasprun"]
else:
    vasprun = Vasprun("vasprun.xml")
    if cached_data is not None:
        cached_data["Vasprun"] = vasprun
```

Any thoughts about this? Would you consider merging a PR implementing this functionality?

shyuep:
I think this is reasonable, though there are a few concerns.

1. Some of the files are changed over the course of a run. In essence, a cached file will miss those changes, esp for handlers that deal with on the fly monitoring. So we have to be careful where we apply this to.
2. I would actually opt for a lru implementation, e.g.,

```
@lru
def load_file(filename):
    ...
    return contents
```

gpetretto:
Hi @shyuep, thanks for looking into this.

Indeed the first point is one to be careful about. My idea was to clear the cached_data dictionary in the `_do_check` method, after all the handlers have been iterated on:
```python
    def _do_check(self, handlers, terminate_func=None):
        corrections = []
        for h in handlers:
            ...
        self.cached_data.clear()
        ...
```
Using the lru_cache would not allow to keep track of which functions have been cached from the Custodian object and prevent from clearing the cache. One option to keep both functionalities would be to implement an additional wrapper to the `@lru_cache` that will also save the set of cached functions, allowing Custodian to clear all the caches when needed. Here a possible implementation:

```python
class tracked_lru_cache:
    cached_functions = set()
    def __init__(self, func):
        self.func = functools.lru_cache()(func)
        functools.update_wrapper(self, func)

    def __call__(self, *args, **kwargs):
        result = self.func(*args, **kwargs)
[...]

PR #273, opened 2023-07-20
Caching parsed output files

gpetretto:
## Summary

Following up on #268 I wrote an initial implementation for caching the output files. It is based on a modified version of the `lru_cache` decorator. This should make it easy switch to helper function for storing different properties, while still allowing `Custodian` to clear the cache after each time the loop on the handlers/validators has been executed.

One potential drawback of the implementation is that the set of cached functions is stored as a class attribute of `tracked_lru_cache`. This means for example that in a multithreading execution (memory shared), with multiple instances of Custodian running, the cache will be cleared more frequently than expected. It should not lead to overlaps, due to the caching being based on the absolute paths. However, a multithreading execution does not seem very likely to be used with Cusodian. No problem will happen in a multiprocessing execution of multiple instances of Custodian (no memory shared).

Making the set of cached function a property of a single Custodian instance is also feasible, but it will need to be passed to the handlers so that the decorated functions can be tracked, making the implementation of the cached functions a bit more involved.

I would like to first have a feedback on this solution. If it is fine I will proceed to write specific tests for the new functionality.

If there is also an interest for switching to the parsing of specific values, instead of the full vasprun.xml/OUTCAR, I will do that in a separate PR.


shyuep:
This generally looks good to me.

gpetretto:
Thanks for looking into this.  
I have added the tests.

janosh:
@gpetretto Looks like this fell under the radar. If you could resolve the conflicts, we'll get this merged!

gpetretto:
Thanks for getting back on this @janosh. It should be fine now.

custodian/vasp/io.py
janosh, line 1: How about instead of a new `io` module, we just define

```py
ChachedOutcar = tracked_lru_cache(Outcar)
ChachedVasprun = tracked_lru_cache(Vasprun)
```

directly in `custodian/vasp/handlers.py`?

custodian/vasp/tests/conftest.py
```diff
@@ -12,3 +12,13 @@ def _patch_get_potential_energy(monkeypatch):
...
+
+
+@pytest.fixture(autouse=True)
+def _clear_tracked_cache():
+    """
+    Clear the cache of the stored functions between the tests.
+    """
+    from custodian.utils import tracked_lru_cache
+
+    tracked_lru_cache.tracked_cache_clear()
```
janosh, line 24: I think this fixture currently runs on _every_ test, not just the handler tests which would be unnecessary work?

janosh: I think you can move it to `custodian/vasp/handlers.py` but keep `autouse=True` [to only auto apply it to those tests](https://docs.pytest.org/en/6.2.x/fixture.html#fixture-availability).

janosh (changes requested):

janosh:
@gpetretto Perhaps you could address my comments in a new PR?

custodian/vasp/io.py
gpetretto: This could be done, but since this is also used in the `validators` module they should be imported from the `handlers`. I think that having a generic io module could be fine, but I have no problem moving them to handlers.py.

custodian/vasp/tests/conftest.py
gpetretto: Since this is also used in the validators, I think the fixture should either be duplicated in the handlers and validators files, or added it explicitly to the tests that need it, removing the `autouse`.

gpetretto:
Hi @janosh , sorry for the delay. I can indeed adress these comments in a separate PR. Can you maybe clarify what would be your preferences in the comments above?

custodian/vasp/tests/conftest.py
janosh: Yeah, probably duplicating it in both files makes it more explicit too. People are more likely to see it and become aware this is happening.

custodian/vasp/io.py
janosh: Yeah, I guess no need to move to `handlers.py`.

PR #275, opened 2023-07-21
`LargeSigmaHandler` set `ISMEAR=0` if sigma reaches `0.02` minimum

janosh:
Half the jobs in our atomate2 r2SCAN trial runs fizzle due to `LargeSigmaHandler`. The failed runs are mostly metals.

After starting with `sigma=0.2` and decreasing 3 times, we reach the minimum of 0.02 allowed by Custodian at which point the `LargeSigmaHandler` aborts the job.

VASP wiki says Gaussian smearing leads to reasonable results in most cases. In the final summary they say if you have no prior knowledge of the system, use `ISMEAR = 0`.

Alternative suggestion from @esoteric-ephemera is to increase the max allowed entropy from 1 meV/atom even though VASP docs recommend. They probably didn't benchmark that value for larger structures. Or better yet, make it a user overridable `entropy_tol` kwarg.

https://github.com/materialsproject/custodian/blob/202b941ab9c5d8fc3b9b9e5bb26a9e280ba1e0ea/custodian/vasp/handlers.py#L1264

### DIFF
--- a/custodian/vasp/handlers.py
+++ b/custodian/vasp/handlers.py
@@ -1287,11 +1287,11 @@ class LargeSigmaHandler(ErrorHandler):
             )
         else:
             # https://vasp.at/wiki/index.php/ISMEAR recommends ISMEAR = 0 if you have
-            # no a priori knowledge of your system ("then always use Gaussian smearing")
+            # no a priori knowledge of your system ("then always use Gaussian smearing"
             actions.append(
                 {
                     "dict": "INCAR",
-                    "action": {"_set": {"ISMEAR": 0}},
+                    "action": {"_set": {"ISMEAR": 0, "SIGMA": 0.05}},
                 }
             )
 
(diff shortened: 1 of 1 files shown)


arosen93:
I agree that ISMEAR = 0 is a good fallback option. It is a safe bet in virtually all cases and is likely to be more robust a solution. However, there is one important caveat. If the user has specified ISMEAR = -5, they almost certainly want to *only* do ISMEAR = -5. Changing ISMEAR = -5 to ISMEAR = 0 would likely cause hidden issues for the user that they likely would not know about. So, my personal $0.02 is to implement a change like this but only if ISMEAR != -5.

janosh:
@arosen93 Looks like the handler only applies if ISMEAR is > 0 to begin with.

https://github.com/materialsproject/custodian/blob/202b941ab9c5d8fc3b9b9e5bb26a9e280ba1e0ea/custodian/vasp/handlers.py#L1254

arosen93:
Beautiful. Then I think this makes sense.

I would also recommend explicitly setting a value for `SIGMA` if you set `ISMEAR = 0`. Perhaps 0.01 or 0.05? Thoughts, @esoteric-ephemera?

esoteric-ephemera:
I second SIGMA = 0.05, that's pretty safe

PR #276, opened 2023-07-24
enable using `check` functions on gzipped files for VASP

matthewkuner:
## Summary

Use methods from `monty` to enable usage of the `check()` functions in Custodian handlers/validators on compressed files. This is (primarily) meant to enable us to use Custodian handlers/validators when validating VASP calculations for the Materials Project Database.


janosh:
Thanks @matthewkuner. It might even make sense to do this in `pymatgen`, i.e. have `Incar.from_file("INCAR")` always look for `INCAR.gz` if `INCAR` is not found. Ofc @shyuep's decision if he wants to weigh in.

shyuep:
I don't have an issue with this of course.

PR #277, opened 2023-07-25
move `algo_tet` changes from `UnrecoverableError` to `VaspErrorHandler` 

janosh:
Also try setting `algo=fast` first, only if that fails try changing smearing.


janosh:
Note for future visitors: We ended up modifying `UnconvergedErrorHandler` in fcd6560 so as not to waste a correction on triggering an unnecessary `algo_tet` error by only changing `ALGO` to `All`/`Normal` if `if v.incar.get("ISMEAR", -1) >= 0`.

PR #283, opened 2023-08-04
Fix AttributeError NoneType has no attribute 'style' 

janosh:
When running VASP with `INCAR` tag `kspacing` instead of a `KPOINTS` file and encountering `brmix` 2 or 3 times. Reported by @esoteric-ephemera in atomate2 r2SCAN workflow.

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -1,6 +1,12 @@
 name: Testing
 
-on: [push, pull_request]
+on:
+  push:
+    branches: [master]
+    paths-ignore: [docs/**]
+  pull_request:
+    branches: [master]
+    paths-ignore: [docs/**]
 
 jobs:
   build:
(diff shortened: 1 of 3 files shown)


janosh:
@arosen93 Would be great if you could take a quick look if this seems sensible.

arosen93:
@janosh: Thanks for the ping. This seems sensible to me.

Are we certain that we can do `if vi["KPOINTS"] is not None`? In other words, we don't need to do the safer option of `if vi.get("KPOINTS")`?

janosh:
Yeah, [looks like the key is always set](https://github.com/materialsproject/pymatgen/blob/c11573882fe9f3774158862309d87a72ee0b98b9/pymatgen/io/vasp/inputs.py#L2439), just initialized to `None` is `KPOINTS` file not found.

But we don't really need if `vi["KPOINTS"] is not None`, could do `if vi["KPOINTS"]`.

PR #284, opened 2023-08-09
Tweak unconverged, bravais and den-/tet error handling

janosh:
225a1e8d UnconvergedErrorHandler only set algo to normal if ISMEAR>=0
608530b2 mv custodian/feff/tests/test_handler{,s}.py
c11ab49e tweak bravais error handling in VaspErrorHandler
283fe9d2 improve den-/tet error handling in case of not using kspacing
bc022852 fix VaspErrorHandlerTest.test_bravais


Andrew-S-Rosen:
@janosh: In the future, can we please not merge in PRs that change handlers in non-trivial ways without it being reviewed by another individual? The changes here to the "tet" handler are not good.

PR #289, opened 2023-09-10
Update @arosen93 to @Andrew-S-Rosen

Andrew-S-Rosen:
I changed by username and did a find-and-replace here.

Related: https://github.com/materialsproject/pymatgen/pull/3311


shyuep:
I think we need a unit test to ensure that arosen93 never appears again. 😂

PR #297, opened 2023-10-19
Fixed a small issue with the erroneous attribute call on a Structure object in AMIN handler

fyalcin:
Recently ran into this issue. `Structure` object doesn't have a `structure` attribute. This is now fixed.

### DIFF
--- a/custodian/vasp/handlers.py
+++ b/custodian/vasp/handlers.py
@@ -1444,7 +1444,7 @@ class NonConvergingErrorHandler(ErrorHandler):
 
         # NOTE: This is the amin error handler
         # Sometimes an AMIN warning can appear with large unit cell dimensions, so we'll address it now
-        if np.max(Structure.from_file("CONTCAR").structure.lattice.abc) > 50.0 and amin > 0.01:
+        if np.max(Structure.from_file("CONTCAR").lattice.abc) > 50.0 and amin > 0.01:
             actions.append({"dict": "INCAR", "action": {"_set": {"AMIN": 0.01}}})
 
         # If a hybrid is used, do not set Algo = Fast or VeryFast. Hybrid calculations do not


janosh:
Thanks @fyalcin! Would you be able to add a test for this?

Andrew-S-Rosen:
Probably best to merge this fix and open an issue marking that we need to test that line.

janosh:
I just noticed we currently have zero test coverage for `NonConvergingErrorHandler`. Sorry @fyalcin, didn't mean to imply you should write a new error handler test from scratch.

@Andrew-S-Rosen @matthewkuner Would either of you know where best to get reference files that match the `NonConvergingErrorHandler.check` conditions?

```py
class NonConvergingErrorHandler(ErrorHandler):
    """
    Check if a run is hitting the maximum number of electronic steps at the
    last nionic_steps ionic steps (default=10). If so, change ALGO using a
    multi-step ladder scheme or kill the job.
    """

    def check(self):
        """Check for error."""
        vi = VaspInput.from_directory(".")
        n_elm = vi["INCAR"].get("NELM", 60)  # number of electronic steps
        try:
            oszicar = Oszicar(self.output_filename)
            elec_steps = oszicar.electronic_steps
            if len(elec_steps) > self.nionic_steps:
                return all(len(e) == n_elm for e in elec_steps[-(self.nionic_steps + 1) : -1])
        except Exception:
            pass
        return False
```

matthewkuner:
Personally, I would just manually modify the file to have extra electronic/ionic steps, or change the parameter for nionic_steps to be lower

fyalcin:
I noticed the lack of error coverage for this particular handler but couldn't find the time to write one myself and decided to go without it. In hindsight, I should have mentioned this in the first message, sorry!

PR #299, opened 2023-11-04
Rewrite handler tests

janosh:
The tests in `custodian/vasp/tests/test_handlers.py` are laden with order-dependent side effects. Running all tests in that file does not leave the `test_files/` unchanged. Large numbers of files are modified and added while tests run and again deleted at the end... but only if all tests pass. Failing tests leave behind dirty files in the repo that then cause subsequent test runs to fail that were previously passing.

The right way to run IO-heavy tests like in Custodian is to move everything into temporary dirs that are auto-cleaned by the OS and can't affect anything inside the repo.

### DIFF
--- a/custodian/vasp/tests/test_handlers.py
+++ b/custodian/vasp/tests/test_handlers.py
@@ -2,7 +2,6 @@
 import datetime
 import os
 import shutil
-import unittest
 from glob import glob
 
 import pytest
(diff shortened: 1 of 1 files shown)


janosh:
Ran out of time. Only ones still missing are `VaspErrorHandlerTest`, `UnconvergedErrorHandlerTest`.

Andrew-S-Rosen:
@janosh: Thank you for doing this valuable cleanup. I had thought about splitting these up into a greater number of individual tests and using pytests's `tmpdir` fixture, but this looks like it works well too.

PR #315, opened 2024-02-23
Add missing init for errors in VASP UnconvergedErrorHandler

esoteric-ephemera:
In a recent PR (#313) I mistakenly moved the init for `errors` in `custodian.vasp.handlers.UncovergedErrorHandler` so that if an error is unfixable, an exception is thrown at when trying to return `{"errors": errors, "actions" : None}`:
```
    return {"errors": errors, "actions": None}
                      ^^^^^^
UnboundLocalError: cannot access local variable 'errors' where it is not associated with a value
```

This PR corrects that initialization and adds a test for uncorrectable, unconverged runs.

Either way, the job would be terminated by custodian, so this isn't an urgent bug. But we want proper tracking of why the job was terminated.


janosh (commented):
can you explain what's happening with `pytest` that requires unzipping the test file in 154fb0f?

esoteric-ephemera:
> can you explain what's happening with pytest that requires unzipping the test file in https://github.com/materialsproject/custodian/commit/154fb0fddc075b8d8cbd400449ad249aaddc3af4?

Not sure - I noticed that if I run one test file specifically, nothing gets unzipped. If I run `pytest tests/` while in the head dir, every file gets unzipped. I'm guessing this is why there's currently a `vasprun.xml.indirect` and `vasprun.xml.indirect.gz` in the test files dir - probably missed after someone ran all tests

esoteric-ephemera:
Also makes sense why the tests fail in CI when I used gzipped files - they get unzipped when `pytest --cov=custodian --color=yes tests` is run

janosh:
> they get unzipped by pytest --cov=custodian --color=yes tests

this is bad! test reference files should not be modified in place. they should be copied to a temp dir first. i refactored most of the handler tests at some point to work that way but looks like the rest of the code base still needs a rework

esoteric-ephemera:
A more significant refactor might be needed for the tests, but might that warrant a separate PR?

janosh:
Definitely, except I would recommend the minimum changes required to keep the new test file added in this PR compressed

PR #317, opened 2024-03-08
Make custodian threadsafe (explicit file paths)

zulissimeta:
Custodian is great, but the expectation is that it's running in its own process and the current working directory is used everywhere. Most file IO is hard-coded to either "./", os.getcwd(), or just implicitly assumes one of those. This could be a problem if multiple threads (say in a workflow) were running custodian jobs, since all threads share the same cwd.

This PR is a first stab at making paths explicitly everywhere so the process cwd is never used implicitly or explicitly. Only the VASP paths have been updated so far.


Andrew-S-Rosen:
Thank you very much for this very PR! I genuinely think this is a very important one to address, and I will be happy to take the lead on getting this over the finish line.

Some things to do:
- Some minor refactoring (e.g. there are some repeated `os.path.join` calls that we can define once and be done with)
- Some simple tests
- Get all checks to pass

shyuep:
Nice. I agree it is needed. Never realized that the relative paths are non-threadsafe.

zulissimeta:
@shyuep It's not obvious to me how to update _mod_input in the nwchem handlers. I don't see it actually used/referenced anywhere; can this be removed? 

https://github.com/materialsproject/custodian/blob/master/custodian/nwchem/handlers.py#L42

shyuep:
@zulissimeta Pls go ahead and delete it. If the tests pass, I have no problem. My philosophy is - if the original author didn't bother to write tests to guarantee a functionality, no one else has a responsibility to ensure that it sticks around.

Also, feel free to just focus on the classes that you actually use (e.g., vasp). There is no need to concern yourself with making other components threadsafe (e.g., nwchem or qchem) unless you are using them or are inclined to do it.

shyuep:
Thanks for all the work!

zulissimeta:
Thanks @shyuep - you might want to leave this open for a little bit; I think @Andrew-S-Rosen was going to take a pass and add some simple tests for the directory functionality.

Andrew-S-Rosen:
Thanks @shyuep and @zulissimeta! I'll open a follow-up PR for additional tests and any fixes.

Issue #316, opened 2024-03-01
[Feature Request]: Add support for a `directory` to run custodian calculations with


Andrew-S-Rosen:
### Problem

Pretty much across the board, Custodian assumes that everything is being done in the local working directory. Oftentimes, this is fine because the user will typically have the calculation run in an isolated location. However, if trying to run calculations in a multi-threaded mode (or any mode where you run multiple calculations per Python process), this can get messy really quickly because you can't call `os.chdir` in such scenarios.

Ideally, the various handlers should take an optional `directory: str | Path = "."` keyword argument, which is then used for all I/O and in the subprocess call to the executable.

### Proposed Solution

Lots of manual insertion of `directory` before filenames and in `subprocess`.

PR #318, opened 2024-03-08
Add support for `directory` for Q-Chem

Andrew-S-Rosen:
This PR directly follows #317 (and addresses #316) by adding a `directory` option for the Q-Chem custodian routines, thereby making it thread-safe and also giving the user flexibility to run beyond the current working directory.


Andrew-S-Rosen:
@janosh --- this should be ready to go. That said, just a head's up that there seems to be a slightly flakey windows test that is seemingly unrelated to this PR.

Sometimes it fails on Windows with [Python 3.10](https://github.com/materialsproject/custodian/actions/runs/8237584828/job/22526658393#step:8:33) and sometimes with [Python 3.12](https://github.com/materialsproject/custodian/actions/runs/8236951251/job/22524586254#step:8:33) and then other times not at all. I have only modified Q-Chem functions, so it doesn't seem to be associated with my changes in any obvious way.

PR #321, opened 2024-03-09
`pip` to `uv` CI migration

janosh:
related PR https://github.com/materialsproject/pymatgen/pull/3675


janosh:
attempted but failed to install `openbabel` with `uv` as well. sticking with slow `conda` for that for now

PR #323, opened 2024-03-11
Add missing `directory` kwarg on QCJob `run()` method

Andrew-S-Rosen:
## Summary

The `QCJob` class' `run()` method was missing a `directory` keyword argument, which is needed in order to have the `subprocess` call run in the pre-specified directory.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -13,11 +13,13 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
+
       - name: Set up Python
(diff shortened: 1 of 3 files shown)


janosh (approved):
would be nice to have a test for `QCJob.run()` but i assume that's hard/slow?

Andrew-S-Rosen:
@janosh: Yeah, it would be nice. I must admit, my familiarity is primarily with the VASP modules, so I'm not sure how viable/not viable it would be to address. I imagine it would just involve a fair bit of monkeypatching.

PR #326, opened 2024-04-01
Enable export of environment variables plus lobster run as a command

JaGeo:
## Summary

This will enable using "NUM_OMP_THREADS=48 /bin/lobster-command" as a command. With this, running VASP and LOBSTER within one job-script with jobflow and atomate2 only (without installing fireworks or jobflow remote) will be possible. If users rely on one 1 CPU node only, this should ensure good run times.

(cc @naik-aakash )

### DIFF
--- a/custodian/lobster/jobs.py
+++ b/custodian/lobster/jobs.py
@@ -87,8 +87,8 @@ class LobsterJob(Job):
             zopen(os.path.join(directory, self.stderr_file), "w", buffering=1) as f_err,
         ):
             # use line buffering for stderr
-            cmd = ' '.join(cmd) # to join split commands (e.g., from atomate and atomate2)
-            return subprocess.run(cmd, stdout=f_std, stderr=f_err, shell=True)  # pylint: disable=R1732
+            cmd = " ".join(cmd)  # to join split commands (e.g., from atomate and atomate2)
+            return subprocess.run(cmd, stdout=f_std, stderr=f_err, shell=True, check=False)  # pylint: disable=R1732
 
     def postprocess(self, directory="./"):
         """Will gzip relevant files (won't gzip custodian.json and other output files from the cluster)."""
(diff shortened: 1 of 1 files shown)


JaGeo:
@janosh would be great if you could have a look at it. Ideally, we have this in the next custodian and  atomate2 version as it would simplify running automated bonding analysis runs quite a lot. 

I changed 2 lines of code here. Any suggestions welcome.

Alternative to this solution: changing both atomate and atomate2 code base to not convert the command into a list. I am a bit hesitant to do that as atomate is not very frequently updated.

custodian/lobster/jobs.py
```diff
@@ -87,7 +87,8 @@ def run(self, directory="./"):
             zopen(os.path.join(directory, self.stderr_file), "w", buffering=1) as f_err,
         ):
             # use line buffering for stderr
-            return subprocess.Popen(cmd, cwd=directory, stdout=f_std, stderr=f_err)  # pylint: disable=R1732
+            cmd = " ".join(cmd)  # to join split commands (e.g., from atomate and atomate2)
```
JaGeo:
```suggestion
            if isinstance(cmd, list):
                cmd = " ".join(cmd)  # to join split commands (e.g., from atomate and atomate2)
```

We could also do this to make it safer in case a string is transferred or replace it with one of the shlex commands:
```suggestion
import shlex
cmd="test"
if isinstance(cmd, list):
    cmd = shlex.join(cmd) 
```

custodian/lobster/jobs.py
```diff
@@ -78,16 +78,16 @@ def setup(self, directory="./"):
 
     def run(self, directory="./"):
         """Runs the job."""
-        cmd = self.lobster_cmd
+        cmd = " ".join(self.lobster_cmd)  # join split commands (e.g., from atomate and atomate2)
```
janosh, line 81: do we want to support both `str` and `Sequence` for `lobster_cmd`?

JaGeo: Yes, I think it would be good. Just made a comment on that with two alternatives. You were faster :D

JaGeo:
Thanks, @janosh !

JaGeo:
Would be great if a new version could be released when the next pymatgen version is released as well.

PR #333, opened 2024-05-09
Improve `LargeSigmaHandler`

esoteric-ephemera:
## Context
The `LargeSigmaHandler` for VASP calculations tries to ensure that the energy contribution to the electronic free energy due to Fermi-level smearing ("smearing entropy") is not in excess of 1 meV/atom. This recommendation comes from [VASP](https://www.vasp.at/wiki/index.php/ISMEAR#How_to_set_ISMEAR).

Currently, this only checks for order >= 1 Methfessel-Paxton smearing (ISMEAR >= 1), probably because the VASP manual doesn't make it clear that Gaussian smearing (ISMEAR = 0) is just order-0 Methfessel-Paxton.

For Fermi-Dirac smearing (ISMEAR = -1), one can simulate the effect of finite electronic temperature, so it's not a good idea to check the entropy in this case.

## Changes
This PR ensures that the smearing entropy is monitored for ISMEAR >= 0, and now looks at the smearing entropy at the end of <i>each</i> ionic step, rather than just the last step.

Additionally, the method by which SIGMA (the smearing width) is reduced has been changed based on the work of [dos Santos and Marzari](https://doi.org/10.1103/PhysRevB.107.195122). (See note below.)

The test for this handler has also been updated.

Copying @matthewkuner who brought this issue up and worked with me to get a fix together

## Minor changes
Adding functionality to gzip test files - pytest seems to unzip them at test time, but *only in CI*. Similar behavior occurs for the Gaussian tests (see the `gunzip_file` function in `tests/gaussian/test_handlers.py`).

The function `get_gzip_or_unzipped` in `conftest.py` is a simple utility to return the gzipped file if it exists, and the unzipped version if that exists (a reverse `monty.os.path.zpath` that also raises an exception if none of the files exist).

## Explanation of new smearing extrapolation
Near the SIGMA = 0, the free energy has a low-order Taylor expansion:
$F(\sigma) \approx E(\sigma \to 0) + \frac{1}{2} \gamma \sigma^2$
where $F(\sigma)$ is the free energy due to smearing, $E(\sigma \to 0)$ is the electronic energy without smearing (the quantity we usually want), and $\gamma$ is related to the second derivative of the smearing entropy wrt $\sigma$.

One can then extrapolate to what the best reduced sigma value $\sigma'$ is by demanding that the smearing energy is $E_s=1$ meV/atom:
$\sigma' = [2 E_s / \gamma]^{1/2} \sigma$.
Practically, we also impose $\sigma' = \max(0.01, 0.8 \cdot [2 E_s / \gamma]^{1/2} \sigma)$ to ensure that $\sigma'$ is not too small. 

The factor of 0.8 helps ensure that the threshold is overshot slightly. The previous `LargeSigmaHandler` decreased SIGMA too slowly in many cases.


Andrew-S-Rosen:
This is slick!!!!

esoteric-ephemera:
Hi @shyuep and @janosh - this should be ready for review whenever you have time. Thanks in advance!

janosh (approved):
looks good to me! thanks @esoteric-ephemera! :

PR #336, opened 2024-06-12
Fix breaking change to ISMEAR=-5 handling for NKPT<4

### LINKED ISSUE #335: [BUG]: Tetrahedron method not properly fixed with 1x1x1 k-points
### Problem

When using ISMEAR=-5 with NKPT<4, VASP crashes with the following error:
```
|     IBZKPT: tetrahedron method fails for NKPT<4. NKPT = 1                   |
```

The solution in this case should be to update ISMEAR to 0 (and set SIGMA to 0.05). However, that's not done.

Instead, Custodian tries to increase the number of k-points by 20% here: https://github.com/materialsproject/custodian/blob/bc9ecbb5c41fbc7a67b880af6cd804cba52ecd05/custodian/vasp/handlers.py#L202-L204

There are two issues with this approach:
1. It just plain doesn't work. If you have 1x1x1 k-points, the 20% increase will change it to 1.2x1.2x1.2, which rounds right back down to 1x1x1. This is a silly exercise. You are only hitting the "tet" error with a small number of k-points, so it's always going to round back down.
2. We really should not be modifying k-points on-the-fly unless this is some last resort. The k-points, like ENCUT, are effectively part of the level of theory being applied here and we should not be changing it so freely. Edit: This is more of a personal opinion, and @esoteric-ephemera has convinced me that we can live with this.

This behavior was a breaking change introduced in https://github.com/materialsproject/custodian/pull/284.

-------

@janosh: I saw you referred to https://vasp.at/forum/viewtopic.php?f=3&t=416&p=4047&hilit=dentet#p4047 in your PR. If you'd like to handle the "dentet" error differently, that's fine, but it should not impact the "tet" error handler in any way. I also am hesitant about updating the number of k-points in general. I would recommend trying to increase NEDOS, as mentioned in the forum, before considering any changes to the number of k-points with the dentet error. In any case, that is secondary to the main issue reported here.

Andrew-S-Rosen:
Closes #335. See the tagged issue report for full details.

In short, this PR reverts the breaking changes made in #284 related to handling ISMEAR = -5 for NKPT<4. The change in #284 does not work and even if it did, it would not be the wisest course of action. I also split up the "tet" and "dentet" error handling because those are two separate errors that can, in principle, be handled differently.

### DIFF
--- a/custodian/vasp/handlers.py
+++ b/custodian/vasp/handlers.py
@@ -193,21 +193,10 @@ class VaspErrorHandler(ErrorHandler):
             # follow advice in this thread
             # https://vasp.at/forum/viewtopic.php?f=3&t=416&p=4047&hilit=dentet#p4047
             err_type = "tet" if "tet" in self.errors else "dentet"
-            if self.error_count[err_type] == 0:
-                if vi["INCAR"].get("KSPACING"):
-                    # decrease KSPACING by 20% in each direction (approximately double no. of kpoints)
-                    action = {"_set": {"KSPACING": vi["INCAR"].get("KSPACING") * 0.8}}
-                    actions.append({"dict": "INCAR", "action": action})
-                elif vi["KPOINTS"] and vi["KPOINTS"].num_kpts < 1:
-                    # increase KPOINTS by 20% in each direction (approximately double no. of kpoints)
-                    new_kpts = tuple(int(round(num * 1.2, 0)) for num in vi["KPOINTS"].kpts[0])
-                    actions.append({"dict": "KPOINTS", "action": {"_set": {"kpoints": (new_kpts,)}}})
-                elif vi["KPOINTS"] and vi["KPOINTS"].num_kpts >= 1:
-                    n_kpts = vi["KPOINTS"].num_kpts * 1.2
-                    new_kpts = tuple([int(round(n_kpts**1 / 3, 0))] * 3)
-                    actions.append(
-                        {"dict": "KPOINTS", "action": {"_set": {"generation_style": "Gamma", "kpoints": (new_kpts,)}}}
-                    )
+            if vi["INCAR"].get("KSPACING"):
+                # decrease KSPACING by 20% in each direction (approximately double no. of kpoints)
+                action = {"_set": {"KSPACING": vi["INCAR"].get("KSPACING") * 0.8}}
+                actions.append({"dict": "INCAR", "action": action})
             else:
                 actions.append({"dict": "INCAR", "action": {"_set": {"ISMEAR": 0, "SIGMA": 0.05}}})
             self.error_count[err_type] += 1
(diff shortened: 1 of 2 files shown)


Andrew-S-Rosen:
@esoteric-ephemera: You may prefer https://github.com/materialsproject/custodian/pull/337 instead since it fixes the handling for "tet" but doesn't alter your changes to "dentet". Let me know your thoughts.

esoteric-ephemera:
@shyuep will do - we're chatting about solutions and we'll post updates here later

Just to get a few initial thoughts down here:
- The tetrahedron method (ISMEAR = -5) will fail with fewer than 4 k-points, that's the `tet` error. `dentet` is similar in that the k-point density is too low to use with tetrahedron method (the Fermi energy can't be determined)
- There are two ways around this:
  - Increase the k-point density to have at least 4 k-points (`tet`) or higher (`dentet`)
  - Change the smearing and hope the k-point density is high enough for that
- We often change k-point density during workflows to reflect the electronic structure (the wider a bandgap, the lower a k-point density a bulk solid needs)

@Andrew-S-Rosen is right that our fix is insufficient. For both cases, the k-point density should be increased until it exceeds 4 k-points (fixes `tet`) ***and*** exceeds the original k-point density

I would avoid changing the smearing method as a last resort, since the energies you get from tetrahedron (fully variational) will necessarily be more accurate than those of e.g., gaussian smearing (non-variational in the total energy, variational in the pseudo-free energy)

Andrew-S-Rosen:
There are some strong counterpoints to be made about modifying k-points, however. Consider one of the most common cases where you get 1x1x1 k-points: a huge structure. If you have something like a MOF with hundreds of atoms per cell, you simply cannot afford to increase the k-points to more than 1x1x1. It would not be logical, both because you'd lose the benefits of the gamma-point only version of VASP and because the cell might be so large that increasing the number of k-points might be a huge computational burden. If we are going to play with modifying the total number of k-points, it needs to be done in such a way that computational cost is taken into account (e.g. by checking lattice parameters, number of atoms, or something).

Regardless, it's clear the "tet" and "dentet" error handling needs to be split up so we can address those differently since only the former is strictly related to NKPT < 4.

esoteric-ephemera:
Sure, for the case of a single k-point in any direction it doesn't always make sense to bump up the k-point density. The safest assumption there is that the user is trying to simulate a lack of PBC in that direction and not adjust it

In the case of KSPACING being too large (for either `tet` or `dentet` errors), it seems clear that the user is doing bulk-like calculations, since this is always automatic grid generation. Decreasing KSPACING is a reasonable first choice then

When using KPOINTS, we can increase the number of k-points on axes where the number of k-points > 1 until the number of k-points (pre-symmetry reduction) is > 4

In the previous two cases, changing smearing would be a fallback

If all three axes have one k-point, then we just change smearing first

Andrew-S-Rosen:
> The safest assumption there is that the user is trying to simulate a lack of PBC in that direction and not adjust it

Just for full clarity, this isn't generally true. Consider a MOF for example. It is a periodic crystal modeled under PBCs like usual, yet you rarely want or need more than 1x1x1 k-point.

In any case, I think your logic seems reasonable on initial read.

Andrew-S-Rosen:
@esoteric-ephemera and I got on a call and have sorted things out. We are going to close this PR in favor of https://github.com/materialsproject/custodian/pull/337. Further review and any changes by @esoteric-ephemera will happen there, although the PR does most of what we ultimately need.

Issue #340, opened 2024-07-02
[Bug]: The corrected INCAR is not stored in the right location when setting a custom `directory`


Andrew-S-Rosen:
In #317 by @zulissimeta, support for a custom `directory` keyword argument was added to the `Custodian` class so that it did not always have to run in the current working directory.

However, it appears this behavior has an important bug with VASP. When a VASP error occurs, Custodian updates the INCAR (as expected). However, this updated INCAR is *not* stored in the specified `directory` but is instead stored in the current working directory. This means that the calculation itself is never actually corrected. It will keep failing since the INCAR remains the same.

Any failing calculation will reproduce this behavior with the following minimal example, assuming your VASP files are stored in a directory `test` and you are running this script from outside of that directory:

```python
from custodian import Custodian
from custodian.vasp.jobs import VaspJob
from custodian.vasp.handlers import (
    FrozenJobErrorHandler,
    IncorrectSmearingHandler,
    KspacingMetalHandler,
    LargeSigmaHandler,
    MeshSymmetryErrorHandler,
    NonConvergingErrorHandler,
    PositiveEnergyErrorHandler,
    PotimErrorHandler,
    StdErrHandler,
    UnconvergedErrorHandler,
    VaspErrorHandler,
    WalltimeHandler,
)

handlers = [VaspErrorHandler(), FrozenJobErrorHandler(), IncorrectSmearingHandler(), LargeSigmaHandler(), MeshSymmetryErrorHandler(), NonConvergingErrorHandler(), PositiveEnergyErrorHandler(), PotimErrorHandler(), StdErrHandler(), UnconvergedErrorHandler(), WalltimeHandler(), KspacingMetalHandler()]

jobs = [VaspJob(["vasp_std"])] # change this to your VASP executable
c = Custodian(handlers, jobs, directory="test")
c.run()
```

For a specific reproduction:

1. Download `reproduction.zip` below
2. Run `unzip reproduction.zip`
3. Add a POTCAR for Cu in `reproduction/test`, like `PAW_PBE Cu 22Jun2005` (not shared here since it is proprietary)
4. Run `python test_custodian.py` using `reproduction` as the working directory.

What you'll see is the following after it runs:

```
/reproduction
--- INCAR <-- this should be in the `test` directory
--- run_custodian.py
--- test
------- all the other files
```

[reproduction.zip](https://github.com/user-attachments/files/16074374/reproduction.zip)

[...]

PR #341, opened 2024-07-02
small fixes to propagate directory

zulissimeta:
## Summary

Address #340 . Problem: the base Modder() class also sets the directory, so the call to the super `__init__` also needs the directory. 

## Todos

Could do with some unit tests! Nothing new added though.


Andrew-S-Rosen:
Hi @shyuep --- whenever you have a moment, would you be willing to mint a release to have this fix on PyPI? Without the patch, Custodian is silently broken if a `directory` is supplied. Thanks!

PR #346, opened 2024-08-17
Create py.typed

Andrew-S-Rosen:
As recommended in [pep 0561](https://peps.python.org/pep-0561/), a blank `py.typed` marker should be included when type hints are used so downstream codes can type check with `mypy` and similar tools. There aren't many type hints in this code, but there are some.

I also removed the reliance on `numpy==1.26.4` in the test suite since this caused other issues due to https://github.com/materialsproject/pymatgen/issues/3990.


Andrew-S-Rosen:
@shyuep: Any idea why a Mac build would fail but Ubuntu would pass? The error seems unrelated to this PR by the way --- but I'm stumped.

PR #355, opened 2024-11-05
Update lobster output file list

naik-aakash:
Hi @JaGeo  and @shyuep, I have updated the list of output files from LOBSTER here. These are optional files that can be generated using lobster >=v5.

This should also get gzipped now if a user runs with keywords that enable such calculations. 

# Todo

- [x] Add test files and update tests


shyuep:
Please have some unittests to make sure that these are dealt with as they should.

naik-aakash:
Hi @shyuep, unit tests have been added now and should be ready to be merged. Happy to address any changes if desired.

Issue #348, opened 2024-09-02
[Bug]: VASP GAMMA version not used when KSPACING is set in INCAR 


naik-aakash:
### What happened?

1. Am trying to run some VASP calculations using atomate2 , where I explicitly set KSPACING in the INCAR
2. Custodian is able to run the VASP job successfully but always uses vasp_std even when KPOINTS resulting from KSPACING is 1x1x1 

# Expected behaviour
vasp_gam version should be used instead of VASP std.

### Version

2024.6.24

### Which OS?

- [ ] MacOS
- [ ] Windows
- [X] Linux

janosh:
did you specify `gamma_vasp_cmd`? it only auto-switches if it finds the vasp gamma executable and the default name it looks for isn't great since afaik it's more commonly named `vasp_gam` than `self.vasp_cmd + ".gamma"`

https://github.com/materialsproject/custodian/blob/880ca8b96ec3adfe57497b4cba1b72242db8be83/src/custodian/vasp/jobs.py#L127-L130

JaGeo:
`KSPACING` is tge important point. We currently don't see how the code is accounting for this.

naik-aakash:
> did you specify `gamma_vasp_cmd`? it only auto-switches if it finds the vasp gamma executable and the default name it looks for isn't great since afaik it's more commonly named `vasp_gam`, not `self.vasp_cmd + ".gamma"`
> 
> https://github.com/materialsproject/custodian/blob/880ca8b96ec3adfe57497b4cba1b72242db8be83/src/custodian/vasp/jobs.py#L127-L130

Hi @janosh , yes, I did set the `gamma_vasp_cmd ` in the atomate2 config file, and atomate2 is able to use the gamma version automatically when I do not set `KSPACING` in the INCAR (this results in the KPOINTS file being explicitly generated). Issue occurs only when `KSPACING` is set.

esoteric-ephemera:
@naik-aakash @janosh: we'd need to add either a handler to check for when the KSPACING-generated k-points is 1x1x1 and centered at $\Gamma$, or use [this formula from the VASP manual](https://www.vasp.at/wiki/index.php/KSPACING) to estimate the k-point grid before running VASP

shyuep:
The VASP gamma code was written in the age before KSPACING was introduced. I am fine with adding logic to check for KSPACING and choosing the gamma version if it results in a 1x1x1 KPOINT. Someone would have to implement it though.

PR #356, opened 2024-11-15
Bugfix in LargeSigmaHandler and expand vasp_gam auto checks

esoteric-ephemera:
Two major changes:
- Since `LargeSigmaHandler` is a monitor handler (checks while output is still being written), it can occasionally misfire when parsing data from OUTCAR. This adds a fix to prevent jobs from being terminated when the handler itself fails to interpret partial file output
- Close [#348](https://github.com/materialsproject/custodian/issues/348) by expanding the scope of k-point checks to include KSPACING, and to also check for grid shifts in KPOINTS

### DIFF
--- a/src/custodian/vasp/handlers.py
+++ b/src/custodian/vasp/handlers.py
@@ -1396,31 +1396,31 @@ class LargeSigmaHandler(ErrorHandler):
             )
             outcar.read_pattern({"completed_ionic_steps": r"(aborting loop)"}, reverse=False, terminate_on_match=False)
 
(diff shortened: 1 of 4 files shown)


naik-aakash:
Thanks @esoteric-ephemera , was about to start working on adding the vasp gamma runs support when using `KSPACING`. Seems already in place 😃

gpetretto:
Hi, we are also getting the same error as in #36. Is there any blocking point to merge this or #361 to solve the issue?

gpetretto:
Thanks!

PR #375, opened 2025-05-03
Use `vasp_gam` for 1x1x1 k-point grids regardless of whether it is gamma-centered or not

### LINKED ISSUE #374: Gamma-point only version of VASP not used unless `Gamma` is in `KPOINTS` file
@esoteric-ephemera:

This discussion focuses on the code block below:

https://github.com/materialsproject/custodian/blob/55ef52e3560109bd98c5af24737190bbe8741b76/src/custodian/vasp/jobs.py#L1008-L1014

If the user runs a calculation with Custodian where the `KPOINTS` file looks like the following, the standard version of VASP will be used even though there is only one k-point. 

```
K-Points
 0
Monkhorst Pack
1 1 1
0 0 0
```

If I understand correctly, I believe that in this case the calculation still is only using the gamma point. As a sanity check, I confirmed that the gamma-point only version of VASP works perfectly fine with such a calculation.

Note that many users likely run a 1x1x1 k-point grid for a large system without setting it to be gamma-centered explicitly. Take Pymatgen's `automatic_density_by_vol` function, which has the `force_gamma` keyword argument set to `False` by default, for instance.

I think we should consider removing the `kpts.style == Kpoints.supported_modes.Gamma` check, but I'd also be interested in your input here.

Andrew-S-Rosen:
## Summary

Closes #374.

This PR makes a change to the logic for choosing whether a VASP calculation should be run with the gamma-point only version of VASP (typically `vasp_gam`) or not. Specifically, we no longer check for whether the KPOINTS file is gamma-centered or not. This is because a 1x1x1 Monkhorst-Pack grid should be the same as a 1x1x1 Gamma-centered grid. An analogous change was made for the KSPACING-related check.

The tests have been updated, and two previously missing tests have been added:
- A test to make sure the standard version of VASP is used when the kpoints are not 1x1x1
- A test to make sure that the standard version of VASP is used when a small KSPACING is set

## Checklist

- [X] Google format doc strings added. Check with `ruff`.
- [X] Type annotations included. Check with `mypy`.
- [X] Tests added for new features/fixes.
- [X] If applicable, new classes/functions/modules have [`duecredit`](https://github.com/duecredit/duecredit) `@due.dcite` decorators to reference relevant papers by DOI ([example](https://github.com/materialsproject/pymatgen/blob/91dbe6ee9ed01d781a9388bf147648e20c6d58e0/pymatgen/core/lattice.py#L1168-L1172))

Tip: Install `pre-commit` hooks to auto-check types and linting before every commit:

```sh
pip install -U pre-commit
pre-commit install
```

### DIFF
--- a/src/custodian/vasp/jobs.py
+++ b/src/custodian/vasp/jobs.py
@@ -1005,11 +1005,7 @@ def _gamma_point_only_check(vis: VaspInput) -> bool:
         # Prevent VASP gamma from being run on DFPT tasks.
         return False
 
-    if (
-        kpts is not None
-        and tuple(kpts.kpts[0]) == (1, 1, 1)
-        and all(abs(ks) < 1.0e-6 for ks in kpts.kpts_shift)
-    ):
+    if kpts is not None and tuple(kpts.kpts[0]) == (1, 1, 1) and all(abs(ks) < 1.0e-6 for ks in kpts.kpts_shift):
         return True
 
     if (kspacing := vis["INCAR"].get("KSPACING")) is not None:
--- a/src/custodian/vasp/jobs.py
+++ b/src/custodian/vasp/jobs.py
@@ -1007,13 +1007,12 @@ def _gamma_point_only_check(vis: VaspInput) -> bool:
 
     if (
         kpts is not None
-        and kpts.style == Kpoints.supported_modes.Gamma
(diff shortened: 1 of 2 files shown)


Andrew-S-Rosen:
All ready now!

shyuep:
Thanks!

PR #396, opened 2025-09-03
Fix Custodian termination behavior when it is running on a different node than the VASP process

Andrew-S-Rosen:
## Summary

This PR overhauls (and greatly simplifies) the termination logic in the `VaspJob`. The major problem this PR seeks to solve is described in https://github.com/Matgenix/jobflow-remote/issues/323#issuecomment-3249870277 and can be summarized as follows.

In some job orchestration setups, Custodian might end up running on a master node with the VASP processes being launched on sister nodes. This is often done, for instance, when requesting a single large Slurm allocation and running many concurrent VASP processes therein.  Currently, Custodian cannot handle this setup, as the Custodian process on the master node seemingly does not have permission to kill the VASP process on the other node(s) in the allocation, and it then defaults to a `killall` command killing everything (including perfectly fine jobs). However, Custodian does have permission to kill the parent process that launches the VASP executable (typically an `srun` or `mpirun` call), which in fact is what the `killall` indiscriminately kills.

This PR resolves the issue and fixes the somewhat hacky logic that was in place before. 

Tagging @MichaelWolloch and @esoteric-ephemera for review, if interested.

I have tested this in three ways:
1. I confirmed that a 1-node job works fine.
2. I confirmed that a 2-node job works fine.
3. I confirmed that a 5-node Slurm allocation running 5 concurrent one-node VASP runs works fine.

Note: The unit tests were written with assistance from Claude because the mocking was a real mess. There were no tests at all before.

My approach is oddly similar to https://github.com/materialsproject/custodian/pull/231, which did not work for multi-node jobs, so I need to confirm that things work okay there.

## Checklist

- [X] Google format doc strings added. Check with `ruff`.
- [X] Type annotations included. Check with `mypy`.
- [X] Tests added for new features/fixes.
- [X] If applicable, new classes/functions/modules have [`duecredit`](https://github.com/duecredit/duecredit) `@due.dcite` decorators to reference relevant papers by DOI ([example](https://github.com/materialsproject/pymatgen/blob/91dbe6ee9ed01d781a9388bf147648e20c6d58e0/pymatgen/core/lattice.py#L1168-L1172))

Tip: Install `pre-commit` hooks to auto-check types and linting before every commit:

```sh
pip install -U pre-commit
pre-commit install
```

### DIFF
--- a/tests/vasp/test_jobs.py
+++ b/tests/vasp/test_jobs.py
@@ -2,7 +2,6 @@ import multiprocessing
 import os
 import shutil
 import subprocess
-import time
 import unittest
 from glob import glob
 from unittest.mock import Mock, patch


shyuep:
Looks good to me.

MichaelWolloch:
Looks very nice. Much less clunky

Andrew-S-Rosen:
@shyuep: After much testing on the clusters here, everything seems to work. Please feel free to merge. We will see if there are any complaints.

PR #410, opened 2025-10-30
Kill VASP on reaching `MaxCorrectionsPerJobError`

janosh:
we noticed that a lot of our nodes had zombie VASP processes that we're hogging compute and memory without doing anything useful after hitting `MaxCorrectionsPerJobError`, preventing new jobs from starting properly (they we're showing as running but not doing anything). it appears that's because `Custodian._run_job` had no logic to actually kill VASP when the default 5 errors per jobs is reached. is this by design? i'm surprised this didn't come up before so maybe we're missing something? fwiw, we're no longer seeing the zombie VASP processes after redeploying with the changes in this PR.

@Andrew-S-Rosen @esoteric-ephemera @shyuep

### DIFF
--- a/src/custodian/custodian.py
+++ b/src/custodian/custodian.py
@@ -456,7 +456,7 @@ class Custodian:
         job.setup(self.directory)
 
         attempt = 0
-        p = None  # Initialize p to None in case the while loop never executes
+        process = None  # Initialize p to None in case the while loop never executes
         while self.total_errors < self.max_errors and self.errors_current_job < self.max_errors_per_job:
             attempt += 1
             logger.info(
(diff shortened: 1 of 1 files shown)


Andrew-S-Rosen:
Looking at your PR, indeed, it does not seem like it would be related (although you never know). I remain just as surprised as you that this behavior would not have been caught before now. I will try to find time to test this in production before it is merged, but it will take me a little bit.

Andrew-S-Rosen:
@janosh --- As a follow-up question, if the VASP job errors out, wouldn't it not be running anymore?

Andrew-S-Rosen:
@janosh: Just a brief update from me. I just used Custodian 2025.10.11 and had a toy job hit the maximum number of errors. Everything aborted as expected, with no lingering VASP process. The relevant logging info is below:

```
custodian.custodian.MaxCorrectionsPerJobError: Max errors per job reached: 2.
```

So, this at least means I did not break anything in my PR as I was inquiring about. But it also means that there are mechanisms in place to properly abort without your proposed changes.

It would be very helpful to get the logging module stdout from the failed run you are referring to, if it happens again.  There is some useful information in here for debugging purposes. Of course, if you are confident that this PR fixes your issue, please let me know. I ran with your PR and confirmed that it at least didn't drastically ruin anything on that same toy run.

PR #412, opened 2025-11-25
Fix `WalltimeHandler` crash on incomplete `OUTCAR` + new test

janosh:
closes https://github.com/materialsproject/pymatgen/issues/2251

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,7 +10,7 @@ repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
     rev: v0.13.3
     hooks:
-      - id: ruff
+      - id: ruff-check
         args: [--fix, --unsafe-fixes]
       - id: ruff-format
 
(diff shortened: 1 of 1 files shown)


shyuep:
I am fine with this. Merged.

tests/vasp/test_handlers.py
```diff
@@ -1016,6 +1016,24 @@ def test_check_and_correct(self) -> None:
...
         os.remove("STOPCAR")
 
+    def test_check_with_malformed_outcar(self, tmp_path: Path) -> None:
+        """Test that WalltimeHandler.check() returns False on malformed OUTCAR.
+
+        This can happen when VASP is mid-write and the file contains incomplete
+        data like a standalone '-' instead of a float. See
+        https://github.com/materialsproject/pymatgen/issues/2251
+        """
+        os.chdir(tmp_path)
```
Andrew-S-Rosen: @janosh: FYI that better practice here would be to do `monkeypatch.chdir(tmp_path)` to not modify the path of the main process in the test suite.

Andrew-S-Rosen: Never mind. I see this is addressed by the `tearDown` call and is done throughout the module.

janosh: strongly agree. been meaning to overhaul the `custodian` test suite for a long time specifically to fix this but no time

PR #414, opened 2025-12-11
Fix Custodian only killing MPI parent process but leaving VASP ghost processes blocking nodes

janosh:
### Problem

When Custodian terminates a VASP job (e.g., due to walltime or error handling), it previously only sent `SIGTERM` to the parent process (`mpirun`/`srun`). This signal often failed to propagate to child `vasp_std` MPI workers, leaving orphaned "ghost" processes that consume CPU/GPU cycles and memory, blocking subsequent jobs from launching altogether or causing them to OOM

**Hypothesis on why this didn't surface before**: we ran into this on thousands of cloud GPUs where process lifecycle differs from traditional HPC clusters. `slurm` automatically cleans up process trees, killing orphaned processes or tasks when a job finishes, ensuring resources are freed. that did not happen in our environment but is Custodian's responsibility since it often kills VASP multiple times (up to 5 by default) in a single job. I think you might see the VASP ghosts even on `slurm` clusters after the first custodian error handler triggers and before the 5th error is reached at which point `slurm` cleans up the whole process tree. this may depend on whether VASP is invoked via `mpirun` or `srun`; if invoked via `srun`, i think it handles process trees on each individual error handler in which case no VASP ghosts. i think `mpirun`spawns processes independently of Slurm’s job step tracking. If VASP crashes, `slurm` may not know about all the children, leaving lingering ghost processes.

### Solution

- Use `os.killpg()` to send signals to the entire process group, ensuring all MPI workers are terminated
- Graceful `SIGTERM` with configurable timeout (default 10s), escalating to `SIGKILL`
- Validate `CONTCAR` (may be empty if job was killed but Custodian would still copy it to `POSCAR`, overwriting atomic positions with blank file) causing next VASP run to fail

### Testing

we deployed this fix to several thousand GPU nodes and it's been running for a week. much lower job failure rate than before
Screenshot 2025-12-03 at 7 12 38 PM

After fix: 0 pods have >1 VASP processes—each GPU runs at most 1 vasp_std, confirming ghost processes are eliminated.

<details><summary>vasp_all_services_analysis</summary>

vasp_all_services_analysis

</details> 

### Changes

- `VaspJob.terminate()`: Kill process group with `os.killpg(SIGTERM/SIGKILL)`
- Add `terminate_timeout` parameter (configurable, default 10s)
- Add `is_valid_poscar()` utility to validate CONTCAR before copy
- Update 9 handlers to check CONTCAR validity
- new unit tests

supersedes #410 cc @Andrew-S-Rosen @esoteric-ephemera @shyuep

### DIFF
--- a/src/custodian/vasp/jobs.py
+++ b/src/custodian/vasp/jobs.py
@@ -740,7 +740,7 @@ class VaspJob(Job):
             try:
                 os.killpg(pgid, signal.SIGTERM)
             except ProcessLookupError:
-                logger.warning(f"Process group {pgid} already dead. No SIGTERM sent.")
+                logger.warning(f"Process group {pgid} not found")
                 return
             except OSError as exc:
                 logger.warning(f"SIGTERM to process group {pgid} failed: {exc}")
@@ -758,7 +758,7 @@ class VaspJob(Job):
             try:
                 os.killpg(pgid, signal.SIGKILL)
             except ProcessLookupError:
-                logger.info(f"Process group {pgid} already dead. No SIGKILL sent.")
+                logger.warning(f"Process group {pgid} not found")
                 return
             except OSError as exc:
                 logger.warning(f"SIGKILL to process group {pgid} failed: {exc}")


Andrew-S-Rosen:
I think this is also going to cause a lot of problems in its current form that we will need to resolve before it is merged. Here are my questions:
- Does the current implementation work appropriately when trying to kill a VASP process that is running on a different node than the Custodian process? See https://github.com/materialsproject/custodian/pull/396 for discussion. I suspect the `os.killpg` will not have permission to kill a process on a different node. This is why we were killing the parent `srun` process.
- Does the current implementation work appropriately when running VASP and correcting a failed run on multiple nodes? See https://github.com/materialsproject/custodian/pull/231, where `os.killpg` was used and did not work with multiple node VASP runs.

P.S. Generally, please restrict PRs to only one bugfix. This codebase is really hard to debug, and including other bugfixes and unrelated changes makes it difficult to go back through the history.

src/custodian/vasp/jobs.py
```diff
@@ -704,19 +710,53 @@ def constrained_opt_run(
...
+                pgid = os.getpgid(pid)
+            except ProcessLookupError:
+                logger.warning(f"Process {pid} not found (already dead)")
+                return
+            logger.info(f"Sending SIGTERM to process group {pgid}")
+            try:
+                os.killpg(pgid, signal.SIGTERM)
+            except ProcessLookupError:
+                logger.warning(f"Process group {pgid} not found (already dead)")
+                return
```
Andrew-S-Rosen: Does this work as expected when VASP is run across multiple nodes? It seems that a similar approach in https://github.com/materialsproject/custodian/pull/231 did not work as expected. I do not know the details offhand.

Andrew-S-Rosen:
@janosh: Would this logic be better? 

```python
def terminate(self, directory="./") -> None:
    """Kill all VASP processes associated with the current job."""
    pid = self._vasp_process.pid

    if self._vasp_process.poll() is not None:
        logger.warning(f"Process {pid} already terminated")
        return

    if os.name != "nt":
        # Look up process group ID
        try:
            pgid = os.getpgid(pid)
        except ProcessLookupError:
            logger.warning(f"Process {pid} not found (already dead)")
            return
    
        # Send SIGTERM to the entire process group
        logger.info(f"Sending SIGTERM to process group {pgid}")
        try:
            os.killpg(pgid, signal.SIGTERM)
            return
        except Exception as e:
            logger.warning(f"Process group {pgid} not terminated: {e}")
        
        # Send SIGKILL to the entire process group
        try:
            os.killpg(pgid, signal.SIGKILL)
            return
        except Exception as e:
            logger.warning(f"Process group {pgid} not killed: {e}")

    try:
        logger.info(f"Killing PID {pid}")
        self._vasp_process.terminate()
        self._vasp_process.wait(timeout=10)
    except subprocess.TimeoutExpired:
        logger.warning(f"Graceful termination did not work. Force killing PID {pid}")
        self._vasp_process.kill()
        self._vasp_process.wait()
```

esoteric-ephemera:
(tagging @tschaume who noticed the high usage on custodian)

PR #437, opened 2026-09-16
Preserve externally updated files when using scratch_dir

### LINKED ISSUE #389: [Bug]: When `scratch_dir` is set, the updated slurm stdout file is overwritten
### Code snippet

```python
import shlex

from custodian.custodian import Custodian
from custodian.vasp.jobs import VaspJob
from custodian.vasp.handlers import VaspErrorHandler

scratch_dir = "/scratch/gpfs/ROSENGROUP/asrosen/custodian_scratch"
full_vasp_cmd = "srun --nodes 1 --ntasks-per-node 112 --exclusive vasp_std"
full_vasp_gamma_cmd = "srun --nodes 1 --ntasks-per-node 112 --exclusive vasp_gam"

split_vasp_cmd = shlex.split(full_vasp_cmd)
split_vasp_gamma_cmd = shlex.split(full_vasp_gamma_cmd)

jobs = [VaspJob(split_vasp_cmd, gamma_vasp_cmd=split_vasp_gamma_cmd)]

c = Custodian(
    handlers=[VaspErrorHandler()],
    jobs=jobs,
    scratch_dir=scratch_dir,
)

c.run()
```

### What happened?

When specifying `scratch_dir` in `Custodian`, it will automatically copy everything from the starting directory to the scratch directory. When the Custodian run is done, it will then copy everything out of the scratch directory back into the starting directory.

https://github.com/materialsproject/custodian/blob/0c1b6a562dd84bfeae66181bf25db72ca9b1c39e/src/custodian/custodian.py#L368-L372

If you launch a Slurm job in `/starting/path`, it will create `/starting/path/slurm-12345.out` and `/starting/path/scratch_link` that links to `scratch_dir`. It will copy `slurm-12345.out` to the `scratch_dir`, but only the original `/starting/path/slurm-12345.out` is updated. When the calculation finishes, the `/starting/path/slurm-12345.out` file is overwritten with the copied file from the start, destroying any logging.

### Version

2025.5.12

### Which OS?

- [ ] MacOS
- [ ] Windows
- [ ] Linux

### Log output

```shell

```

CAOShurong:
## Summary

- track files in the original working directory while a run executes in `scratch_dir`
- keep scheduler or other externally created, updated, or removed files from being overwritten by the scratch copy-back
- preserve the installed Monty version's existing copy-back behavior for scratch-controlled files

## Testing

- `uv run --frozen pytest tests/test_custodian.py -v` (11 passed)
- focused regression on Python 3.13 with Monty 2026.7.16 (passed)
- `uv run --frozen pre-commit run --files src/custodian/custodian.py tests/test_custodian.py`
- full Windows suite: 106 passed, 28 skipped, with the existing CP2K test unable to launch the Unix `echo` executable under `shell=False`

Closes #389

Disclosure: This contribution was developed with OpenAI Codex assistance. I reviewed and validated the implementation and tests.


Andrew-S-Rosen:
Thank you for the contribution. However, it seems like this change should be committed as a keyword argument to `ScratchDir` in monty rather than to Custodian directly since you are effectively modifying the behavior of `ScratchDir`. Custodian would then call that class with the updated keyword argument.

shyuep:
> 🤖 **Automated PR review** generated by Claude on behalf of @shyuep. Not a human review — treat as a first pass.

The approach — snapshot `(mode, dev, ino, size, mtime_ns, ctime_ns)` of the original directory on enter, and on exit drop from the scratch tree anything whose original-directory state changed — is a sound fix for #389, and it handles both the current Monty copy-back (`exit_copy` + mtime warning + `copytree`) and the legacy `delete_removed_files` path. The test covers the four relevant cases (externally appended, externally created, externally removed, scratch-updated).

**CI:** only pre-commit.ci has run. The `Test` and `Lint` GitHub Actions runs are in `action_required` (first-time contributor) and need a maintainer to approve them before this can be judged green.

**Main concern — silent conflict resolution.** If a file is modified both externally *and* by the job in scratch (or removed externally while the job rewrote it), the scratch version is discarded with no signal. Because the scratch copy is deleted before `super().__exit__()`, Monty's own `RuntimeWarning` for newer-in-CWD files no longer fires either, so this PR removes the one diagnostic that existed. Please `logger.warning` (or `warnings.warn`) the paths preserved from the original directory in favour of the scratch copy. For the `slurm-*.out` case this is noise-free; for a real collision it is the only way a user finds out.

**Secondary**
- `_ScratchDir` couples to Monty internals via `getattr(self, "end_copy", getattr(self, "exit_copy", ...))` and `hasattr(self, "delete_removed_files")`. I cannot find `end_copy` in any Monty release, and `delete_removed_files` is now a deprecated no-op. Since `pyproject.toml` pins `monty>=2.0.6`, either bump the pin to a version with the current `exit_copy`/`pass_through` attributes and drop the legacy branch, or add a comment stating which Monty versions each branch targets.
- Two full recursive `os.walk` + `lstat` passes over the original directory (enter and exit). Fine for a typical VASP run directory; worth a one-line comment that this is O(files), so nobody later points `Custodian` at a large tree unaware.
- Test nit: `monkeypatch.setattr("monty.tempfile.os.symlink", ...)` patches `os.symlink` process-wide for the duration of the test (`monty.tempfile.os` is the `os` module). If the intent is Windows compatibility, gate it on `sys.platform`; on POSIX the real symlink should just be created and cleaned up by `ScratchDir`.

Andrew-S-Rosen:
I don't have qualms with AI-generated code, but seeing as the contributor does not seem to work in the space (please correct me if I'm wrong), I am not confident that this is as thoroughly tested as we would like. This code touches major areas of the Custodian codebase, and I'd feel a lot more reassured if someone actually used this for a real simulation task. 

That said, I still think monty is the best place for the core of this update otherwise. At this time, I do not approve of the PR for merging.

Issue #398, opened 2025-09-04
[Bug]: Custodian cannot be run safely when the executable is on a separate node


Andrew-S-Rosen:
### Code snippet

```python

```

### What happened?

For full context of this issue, refer to the summary in https://github.com/materialsproject/custodian/pull/396.

> Custodian might end up running on a master node with the VASP processes being launched on sister nodes. This is often done, for instance, when requesting a single large Slurm allocation and running many concurrent VASP processes therein.  Currently, Custodian cannot handle this setup, as the Custodian process on the master node seemingly does not have permission to kill the VASP process on the other node(s) in the allocation, and it then defaults to a `killall` command killing everything (including perfectly fine jobs). However, Custodian does have permission to kill the parent process that launches the VASP executable (typically an `srun` or `mpirun` call), which in fact is what the `killall` indiscriminately kills.

#396 solves this for VASP, but essentially the same problem exists for the other codes. The fix in #396 is quite easy to implement for other codes once it is merged.

### Version

2025.8.13

### Which OS?

- [ ] MacOS
- [ ] Windows
- [ ] Linux

### Log output

```shell

```

PR #438, opened 2026-09-16
Fix CP2K termination to avoid host-wide killall

CAOShurong:
## Summary

- retain the exact CP2K launcher returned by `Popen` and start it in an isolated process session
- terminate only that job's POSIX process group, escalating from `SIGTERM` to `SIGKILL` after a configurable timeout
- fall back to terminating the launcher on Windows instead of issuing a host-wide `killall`

This addresses the CP2K portion of #398.

## Why

`Cp2kJob.terminate()` currently runs `killall` for any command element containing `cp2k`. On a shared node, that can terminate unrelated calculations, and it does not reliably target the `srun` or `mpirun` launcher that Custodian actually owns.

The new flow mirrors the proven VASP job lifecycle: `run()` retains the launcher handle, POSIX jobs receive their own process group, and termination is scoped to that group with bounded escalation.

## Testing

- regression first failed because `run()` neither stored the launcher nor created a new session
- 7 focused CP2K job tests pass, covering process retention, already-finished jobs, POSIX `SIGTERM`, timeout escalation, Windows fallback, and force-kill fallback
- changed-file pre-commit passes: Ruff check/format, Mypy, Pyright, whitespace and repository hygiene hooks
- a real Windows subprocess launched through `Cp2kJob.run()` was observed running and then terminated through the fallback path
- `git diff --check` passes

## Checklist

- [x] Google-format docstrings added
- [x] Type annotations included
- [x] Tests added for the fix
- [x] DCO sign-off included

Implementation and test assistance used OpenAI Codex; I reviewed the resulting diff and verification output.


shyuep (commented):
**Automated PR review generated by Claude** (posted on behalf of @shyuep by a scheduled routine; static analysis of the diff + CI only, no code executed locally — treat as advisory).

Good change. This ports the `VaspJob` process-group lifecycle (`start_new_session=True`, `killpg` SIGTERM→SIGKILL with bounded `wait`, launcher fallback) to `Cp2kJob` essentially verbatim, which is the right way to remove the host-wide `killall`. The mocked tests cover every branch (already-finished, SIGTERM, escalation, Windows fallback, kill fallback).

Points to address:

1. `self._cp2k_process` is only assigned inside `run()`. Calling `terminate()` before `run()`, or after a `run()` that raised inside `Popen`, now raises `AttributeError` where the old code was a no-op. `VaspJob` has the same gap, so either initialise `self._cp2k_process = None` in `__init__` and early-return with a warning in `terminate()`, or accept the inconsistency knowingly. The former is cheap.
2. The ~60-line `terminate()` body is now duplicated between `VaspJob` and `Cp2kJob`. A shared `_terminate_process_group(proc, timeout)` helper (in `custodian/utils.py` or on the `Job` base) would remove the drift risk. Fine as a follow-up if you want to keep this PR minimal.

CI: only pre-commit.ci ran (green). The GitHub Actions test matrix (ubuntu/windows × 3.10/3.13) has not executed on this branch — first-time-contributor workflow approval is needed before merge.

PR #439, opened 2026-09-19
Write STOPCAR before VASP SIGTERM so vasprun.xml can flush

shaneraphel:
`VaspJob.terminate` treats `directory` as unused and sends SIGTERM then SIGKILL. Monitor handlers with `is_terminating=True` call that path while VASP is still writing `vasprun.xml`. The recovered INCAR correction then fails `VasprunXMLValidator` on a truncated tree. Wall-time and checkpoint handlers already use `is_terminating=False` plus `LSTOP = .TRUE.` so VASP flushes itself.

`terminate` now writes STOPCAR when none is present, waits up to `terminate_timeout`, and keeps SIGTERM→SIGKILL if VASP is still alive.

Fixes #425


Andrew-S-Rosen:
PRs submitted by those who have no familiarity with Custodian or intention to use it will not be reviewed. If you feel this is not the case for you, please let me know.

